PR #61, opened 2017-08-13
Pushover integration 

marshyonline:
Have added integration for [Pushover](https://pushover.net/)

An application key is required and can be obtained for free from the pushover website.
Needs to be enabled in settings and then enabled for each alias to be sent via the service.


marshyonline:
Before merging, should I add a setting to change the sound and priority of the message?

davidmckenzie:
Looks good mate - only concern is the need for table structure change on upgrade. Will test and do up a run sheet similar to the last release. :)

Re sound and priority - couldn't hurt to have it in the settings, unless that's configurable from the pushover side. I haven't played with pushover much, so not sure how much flexibility you've got there. Probably wouldn't want to store it on a per-alias level at this point, since it might slow some of those precarious SQL queries 😉

marshyonline:
All that is needed is one extra column in the capcodes db, set to 0 as default.
Im in the process of making this query now.
Have just moved the version to my prod env, will see how it handles.

Should and priority would be a global setting and stored in the config

marshyonline:
It should be noted as well, on a new DB the table is already setup by default with the added column

marshyonline:
ALTER TABLE capcodes ADD COLUMN push INTEGER DEFAULT 0;

This looks like it has done the trick,
You might have a better way, SQL never liked me haha 

There is no extra SQL queries added in, i have used the existing query to pull the extra piece of data from the same row.

davidmckenzie:
Ah sweet, it's only _modifying_ columns that needs the whole create/rename table malarkey, adding is fine with alter table statement.

marshyonline:
That's Sound and priory support added, I don't see much besides adding to list of features that needs to be done

davidmckenzie:
Looks good dude! Will merge and add the upgrade notes on the release :)

Thanks!

Issue #66, opened 2017-10-04
Pushover Priority Per Alias


thewillhargan:
Got the program up and running and it's great, but I'd really like to be able to push different aliases at different priorities.

For Example. I'd like to be able to push any messages that would usually go to my pager at Emergency Priority. Whereas other pages can be sent at normal priority

davidmckenzie:
Would need DB structure change unfortunately - will have to keep this one on the backlog for when I eventually get around to redoing the DB side of things. Will be looking at backing it on MongoDB or DynamoDB :)

marshyonline:
This is something i never thought of when i integrated pushover initially, but i really do like this idea. 
But David is right, its not something i would want to push out as it will require another database change and that's something I want to avoid doing again.
+1 on this idea, keen to see it in the newer versions

marshyonline:
I have made this change in #74 :)

PR #74, opened 2018-04-23
Email/SMTP intergration

marshyonline:
# Built in email support!

I have built in email support much like PDW.
New users need to make no DB changes, anyone updating from <=0.1.4 will need to run the following:

`ALTER TABLE capcodes ADD COLUMN mailenable INTEGER DEFAULT 0;`
`ALTER TABLE capcodes ADD COLUMN mailto TEXT;`
`ALTER TABLE capcodes ADD COLUMN pushpri TEXT;`

Supports SSL/TLS
Global on/off option in settings.
Per alias on/of
The TO address is set per alias and allow multiple recipients(Comma separated).

This update bumps the version to 0.1.5

Leverages the same query much like my pushover modification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ The UI is built around a Node/Express/Angular/Bootstrap stack, while the client 
 * Pretty HTML5
 * Native browner notifications
 * [Pushover](https://pushover.net/) Integration - near realtime muti-device notification service
+* SMTP/Email support
 * May or may not contain cute puppies
 
 ### Planned Features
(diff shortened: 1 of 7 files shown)


marshyonline:
Before we merge this, I'm working on #66 
Two birds with one stone, seeing how the DB update could be going out anyway.

marshyonline:
Latest commit apply's the suggested change in #66.
Have changed original pull req to include the db modification for it too

kiwiscout:
Scratching my head wondering why the web service wouldn't restart. Then realised you can run node app.js.. oh look. Needed to add PM2 package 'nodemailer' to get it to start :D

davidmckenzie:
Whoops :D

Updated the release notes to mention this. Thanks @kiwiscout!

PR #79, opened 2018-07-06
Fix for undefined body in response

marshyonline:
API returns a 'undefined' body response to the client because it is unable to see the lastID
Quick and dirty but works.


marshyonline:
That is correct, but looking at it, if it failed it should send a 500 with the error?

marshyonline:
Yeah it looks to be the case :( 
I'm not the sharpest tool in the shed so I could be missing something.

PR #83, opened 2018-07-28
Fix for Multimon-NG Square bracket encoding

DanrwAU:
Hi David,

Made some tweaks to the reader script to handle square brackets common in Victorian pager messages.
Multimon uses German charsets which spits out Ä and Ü characters instead of the brackets.

Looks like multimon-ng gave up on fixing this issue EliasOenal/multimon-ng#69

Also changed the console output to the padAddress just because the alignment was being messed up.

I don't really know much about node so these may cause issues in other places but i have been running this for a month with no issues.

Cheers
Dan

P.s - Sorry about the spam - realized i had missed a push with the correct code.


pyrog:
See also https://github.com/EliasOenal/multimon-ng/issues/69

PR #86, opened 2018-08-15
Add Pushover Per Alias Settings. 

DanrwAU:
Changed Pushover Notification Sound and User/Group keys to be per alias settings rather than global as suggested in https://github.com/davidmckenzie/pagermon/issues/85

This will require a database change unfortunately. Bumped the version number to 0.1.6 due to this. 

`ALTER TABLE capcodes ADD COLUMN pushgroup TEXT;
ALTER TABLE capcodes ADD COLUMN pushsound TEXT;`

Would like if you guys could review, as i hacked this together without really knowing what i'm doing - it seems to work though!

### DIFF
--- a/server/routes/api.js
+++ b/server/routes/api.js
@@ -367,7 +367,9 @@ router.get('/capcodes/:id', function(req, res, next) {
                         "agency": "",
                         "icon": "question",
                         "color": "black",
-                        "push" : "",
+                        "push": "",
+                        "pushgroup": "",
+                        "pushsound": "",
                         "pushpri": "0",
                         "mailenable" : "",
                         "mailto" : ""
@@ -401,7 +403,9 @@ router.get('/capcodeCheck/:id', function(req, res, next) {
                         "agency": "",
                         "icon": "question",
                         "color": "black",
-                        "push" : "",
+                        "push": "",
+                        "pushgroup": "",
+                        "pushsound": "",
                         "pushpri": "0",
                         "mailenable" : "",
                         "mailto" : ""
@@ -456,8 +460,6 @@ router.post('/messages', function(req, res, next) {
         var pdwMode = nconf.get('messages:pdwMode');
         var pushenable = nconf.get('pushover:pushenable');
         var pushkey = nconf.get('pushover:pushAPIKEY');
-        var pushgroup = nconf.get('pushover:pushgroup');
-        var pushSound = nconf.get('pushover:pushSound');
         var mailEnable = nconf.get('STMP:MailEnable');
         var MailFrom      = nconf.get('STMP:MailFrom');
         var MailFromName  = nconf.get('STMP:MailFromName');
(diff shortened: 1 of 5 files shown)


server/routes/api.js
```diff
@@ -484,11 +486,13 @@ router.post('/messages', function(req, res, next) {
                         res.status(200);
                         res.send('Ignoring duplicate');
                     } else {
-                        db.get("SELECT id, ignore, push, pushpri, mailenable, mailto FROM capcodes WHERE ? LIKE address ORDER BY REPLACE(address, '_', '%') DESC LIMIT 1", address, function(err,row) {
+                        db.get("SELECT id, ignore, push, pushpri, pushgroup, pushsound, mailenable, mailto FROM capcodes WHERE ? LIKE address ORDER BY REPLACE(address, '_', '%') DESC LIMIT 1", address, function(err,row) {
                             var insert;
                             var alias_id = null;
                             var pushonoff = null;
                             var pushpri = null;
+                            var pushgroup = null;
```
davidmckenzie: There's a bit of a weird mix of casing here and 508/509. I see it was present in the previous commits for the pushover stuff, so probably all good. I don't have pushover so can't test properly - as long as it's all good for you then I'll merge, just wanted to highlight in case it was unintentional :)

server/routes/api.js
```diff
@@ -456,8 +460,6 @@ router.post('/messages', function(req, res, next) {
         var pdwMode = nconf.get('messages:pdwMode');
         var pushenable = nconf.get('pushover:pushenable');
         var pushkey = nconf.get('pushover:pushAPIKEY');
-        var pushgroup = nconf.get('pushover:pushgroup');
-        var pushSound = nconf.get('pushover:pushSound');
```
davidmckenzie: For backwards compatibility it might be wise to retain this - maybe keep the config file vars, have a test for if the group/sound columns are non-empty in the DB row then use those, otherwise use the defaults specified in the config?

server/routes/api.js
DanrwAU: I'm shocking at this - I'm used to working in powershell where case doesn't matter ;)

From memory i matched it to the casing of the rest of the pushover stuff as you said.

server/routes/api.js
DanrwAU: It *seems* to not cause any issues leaving them out - i've copied my production database over to my test one running this version without issue. 
The only problem will be that pushover will need to be reconfigured on the alias level to work after upgrading. 

I thought about writing in a way for it to use the values set on the settings screen if there was nothing set at the alias level, but then i realized i have no idea how to do that ;)

Might be something for me to do over the weekend.....

marshyonline:
I'll take a look over this today, Just wanna spin it up locally first :)

davidmckenzie:
Merging as is, thanks guys!

PR #87, opened 2018-08-16
Add ability to filter capcodes and source from main index

DanrwAU:
This will require the below to be added to the users config.json file as per the changes to default.json.

Haven't quite got a grip on how to handle that yet. 

This partially fixes https://github.com/davidmckenzie/pagermon/issues/38 these are still visible in the api...

<pre>"HideCapcode": false,
"HideSource": false</pre>

Thanks to @damiankw for assisting with this


marshyonline:
This :+1:

PR #88, opened 2018-08-17
Open Alias Edit screen in new tab/window

DanrwAU:
Modifies the home page to open the New/Edit Alias screen in a new window as a workaround for https://github.com/davidmckenzie/pagermon/issues/47

This behavior probably isn't desirable for the filter clickables (Source/Agency/Capcode) as a majority of users would want to be filtering in the window they are in.

### DIFF
--- a/server/views/index.ejs
+++ b/server/views/index.ejs
@@ -441,9 +441,10 @@ var app = angular.module('app', ['ngRoute', 'ngResource', 'ngCookies', 'angular-
       };
       
       $scope.editAlias = function(alias){
-        // admin urls need to use $window since they don't use this route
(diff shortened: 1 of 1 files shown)


DanrwAU:
I've investigated ways to preserve Middle Click and Right Click for the Filters and can only come up with ways that make the characters clickable, not the entire table row around them. 
May be more trouble than it is worth or that could be a compromise worth the extra functionality. Let me know if you want to see an example 👍

davidmckenzie:
Good call, I did have a half-baked idea on how to solve this but one of the many things I never got around to :)

Was thinking to have the clicking action open a popover with real links in there to edit, filter/search, etc.

PR #96, opened 2018-09-23
Event fixes

marshyonline:
Increases the max events :D

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -100,10 +100,10 @@ var io = require('socket.io').listen(server);
 io.sockets.on('connection', function (socket) {
     socket.removeAllListeners();
     debug('client connect to normal socket');
-    socket.on('echo', function (data) {
-        io.sockets.emit('message', data);
-        console.log('message', data);
-    });
+//    socket.on('echo', function (data) {
+//        io.sockets.emit('message', data);
+//        console.log('message', data);
+//    });
 });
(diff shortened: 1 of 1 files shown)


davidmckenzie:
Weird that the setMaxListeners would kill it like that... wonder if it has to be called further down or something. Could be we need to call it separately for each socket:

e.g.
```
io.sockets.on('connection', function (socket) {
    socket.setMaxListeners(20);
```

Does it show up in browser dev tools as trying to connect to the admin socket?

marshyonline:
Im able to see the socket opening up on the client side.
You can see it by checking the frames of the web socket request.

This site is in HideCapcode mode and they seem to be coming over the admin socket ok

image

image

DanrwAU:
Interesting.... It was 2am thanks for checking on your end.

I'll have a look later today and see if I can figure it out :smile:

marshyonline:
Very strange, this was prior to the pushover fixes pulled into master today.
Ill run the update tonight and ensure something in there is broken :)

DanrwAU:
It very well may be :confused:

DanrwAU:
Interesting... 

I can see them being emitted to the socket - but the page is not refreshing....

image

image

DanrwAU:
I quit.. 

Repulling it down from the merged one fixed it .... ignore me i guess 😠

davidmckenzie:
My money's on caching - the client side JS was probably cached, so it didn't know what to do with the events

PR #103, opened 2018-10-14
Add GUI Validation for pushover key

DanrwAU:
Finishes completing https://github.com/davidmckenzie/pagermon/issues/94

Add's GUI Check to prevent saving with a blank or 0 key.  

@marshyonline @davidmckenzie  if you wouldn't mind checking my work on this one, haven't done alot with this file before, seems to work fine on my test box 😄


DanrwAU:
Still in 2 minds about this - it probably should be a validation in api.js now that i think about it.... thoughts?

davidmckenzie:
Logic seems sound, but wonder if we could just handle it as a required field when pushover is true. We should probably grey out those fields too when the pushover checkbox is unchecked. I'll have a poke and see if I can come up with something.

PR #110, opened 2018-10-27
Handle Fragmented messages

Shaggs:
I know its not pretty but not bad for first attempt.

It looks for the frag maker if F (fragment) is found its stored until C is found (Cont) and joins them together. if K is found it just continues as usual

### DIFF
--- a/client/reader.js
+++ b/client/reader.js
@@ -74,18 +74,26 @@ rl.on('line', (line) => {
     }
   } else if (line.indexOf('FLEX: ') > -1) {
     address = line.match(/FLEX:.*?\[(\d*?)\] /)[1].trim();
-    if (line.match( /( ALN | GPN | NUM)/ )) {
-      message = line.match(/FLEX:.*?\[\d*\] ... (.*?)$/)[1].trim();
-      trimMessage = message;
-    } else {
-      message = false;
-      trimMessage = '';
+      if (line.match( /1600\/2\/F\/A/ )) {
+       if (line.match( /( ALN | GPN | NUM)/ )) {
+        frag = line.match(/FLEX:.*?\[\d*\] ... (.*?)$/)[1].trim();
+        message = false;
+        trimMessage = '';
+       }
+      }
+      if (line.match( /1600\/2\/C\/A/ )) {
+       if (line.match( /( ALN | GPN | NUM)/ )) {
+          message = line.match(/FLEX:.*?\[\d*\] ... (.*?)$/)[1].trim();
+          trimMessage = frag+message;
+          }
+        }
+      if (line.match( /1600\/2\/K\/A/ )) {
+        if (line.match( /( ALN | GPN | NUM)/ )) {
+        message = line.match(/FLEX:.*?\[\d*\] ... (.*?)$/)[1].trim();
+        trimMessage = message;
+        }
+       }
     }
-  } else {
-  	address = '';
-  	message = false;
-  	trimMessage = '';
-  }
 
   // filter out most false hits
   // if too much junk data, make sure '-p' option isn't enabled in multimon
--- a/client/reader.js
+++ b/client/reader.js
@@ -96,10 +96,10 @@ rl.on('line', (line) => {
         trimMessage = message;
       }
     }
-  } else {	
-    address = '';	
-    message = false;	
-    trimMessage = '';	
+  } else {
+    address = '';
+    message = false;
+    trimMessage = '';
   }
 
   // filter out most false hits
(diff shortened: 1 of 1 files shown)


davidmckenzie:
Sorry dude, keep meaning to look at this then get pulled away to other things :) I'll push a few changes soon, but this is pretty awesome for first touch of node 😁

client/reader.js
```diff
@@ -74,18 +74,26 @@ rl.on('line', (line) => {
...
+        message = line.match(/FLEX:.*?\[\d*\] ... (.*?)$/)[1].trim();
+        trimMessage = message;
+        }
+       }
     }
-  } else {
-  	address = '';
-  	message = false;
-  	trimMessage = '';
-  }
```
davidmckenzie: What was this else block removed for? Doesn't look like it's covered elsewhere. IIRC, this bit was needed for when multimon does weird stuff.

client/reader.js
```diff
@@ -74,18 +74,26 @@ rl.on('line', (line) => {
...
     address = line.match(/FLEX:.*?\[(\d*?)\] /)[1].trim();
-    if (line.match( /( ALN | GPN | NUM)/ )) {
-      message = line.match(/FLEX:.*?\[\d*\] ... (.*?)$/)[1].trim();
-      trimMessage = message;
-    } else {
-      message = false;
-      trimMessage = '';
+      if (line.match( /1600\/2\/F\/A/ )) {
+       if (line.match( /( ALN | GPN | NUM)/ )) {
+        frag = line.match(/FLEX:.*?\[\d*\] ... (.*?)$/)[1].trim();
```
davidmckenzie: So frag needs to be declared further up - should be declared above line 54 (outside the `rl.on`). If it's declared within the `rl.on` block then it'd be overwritten on every line. Just adding `var frag;` up around line 40 should be all that's needed.

client/reader.js
```diff
@@ -74,18 +74,26 @@ rl.on('line', (line) => {
...
   } else if (line.indexOf('FLEX: ') > -1) {
     address = line.match(/FLEX:.*?\[(\d*?)\] /)[1].trim();
-    if (line.match( /( ALN | GPN | NUM)/ )) {
-      message = line.match(/FLEX:.*?\[\d*\] ... (.*?)$/)[1].trim();
-      trimMessage = message;
-    } else {
-      message = false;
-      trimMessage = '';
+      if (line.match( /1600\/2\/F\/A/ )) {
+       if (line.match( /( ALN | GPN | NUM)/ )) {
```
davidmckenzie: This if is repeated in each of the three `/1600\/2\/F\/A/` ifs, I'd suggest moving it one level up instead, e.g.

```
if (line.match( /( ALN | GPN | NUM)/ )) {
  if (line.match( /1600\/2\/F\/A/ )) {
    frag = line.match(/FLEX:.*?\[\d*\] ... (.*?)$/)[1].trim();
    message = false;
    trimMessage = '';
  }
  else if (line.match( /1600\/2\/C\/A/ )) {
  ...
  }
  ...
}
```

client/reader.js
```diff
@@ -74,18 +74,26 @@ rl.on('line', (line) => {
     }
   } else if (line.indexOf('FLEX: ') > -1) {
     address = line.match(/FLEX:.*?\[(\d*?)\] /)[1].trim();
-    if (line.match( /( ALN | GPN | NUM)/ )) {
-      message = line.match(/FLEX:.*?\[\d*\] ... (.*?)$/)[1].trim();
-      trimMessage = message;
-    } else {
-      message = false;
-      trimMessage = '';
+      if (line.match( /1600\/2\/F\/A/ )) {
```
davidmckenzie: For these, I'd probably want to match it a little wider in case people have different baud rates etc.

```suggestion
      if (line.match( / [0-9]{4}\/[0-9]\/F\/. / )) {
```

This should be enough to match the pattern without false positives. I've also added spaces on either side of the string, to further reduce false positives

davidmckenzie:
Oh, just had a thought - need to include a case for people on the pre 0.8.7 version of multimon, where the KFC key doesn't exist, otherwise this would just ignore messages for them :)

Shaggs:
I think making the latest version of multimon a prerequisite would be best.  Otherwise fragmented messages can't be handled at all. Other option is make a branch for older versions of multimon.

Shaggs:
@davidmckenzie are you able to make the changes as this goes a bit beyond my scope :)
For the first time since running this on my PC ive had only one fragmented message not marry up correctly or close to what it should be +/- a few letters. i think this will be the best we can do

davidmckenzie:
Yeah all G dude, I'll put something in for this on the weekend probably, just a little time poor at the moment :)

PR #113, opened 2018-11-01
Add Telegram Support, Add Automated Database upgrade - Bump to version 0.1.9

DanrwAU:
Adds Telegram support as a notification provider.

Requires the following on existing databases

```
ALTER TABLE capcodes ADD COLUMN telegram INTEGER DEFAULT 0; 
ALTER TABLE capcodes ADD COLUMN telechat TEXT;
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,10 @@ The UI is built around a Node/Express/Angular/Bootstrap stack, while the client 
 * WebSockets support - messages are delivered to clients in near realtime
 * Pretty HTML5
 * Native browner notifications
-* [Pushover](https://pushover.net/) Integration - near realtime muti-device notification service
-* SMTP/Email support
+* Push notifications
+    * [Pushover](https://pushover.net/) Integration - near realtime muti-device notification service
+    * [Telegram](https://telegram.org/) Integration - near realtime cloud based multi-device messaging
+    * [Email] Support for conventional SMTP email notifications 
 * May or may not contain cute puppies
 
 ### Planned Features
@@ -33,7 +35,6 @@ The UI is built around a Node/Express/Angular/Bootstrap stack, while the client 
 * Bootstrap 4 + Angular 2 support
 * Enhanced alias control
 * Graphing
-* Push notifications
 * Non-sucky documentation
 
 ### Screenshots
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The UI is built around a Node/Express/Angular/Bootstrap stack, while the client 
 * Push notifications
     * [Pushover](https://pushover.net/) Integration - near realtime muti-device notification service
     * [Telegram](https://telegram.org/) Integration - near realtime cloud based multi-device messaging
-    * [Email] Support for conventional SMTP email notifications 
+    * Email Support for conventional SMTP email notifications 
 * May or may not contain cute puppies
 
 ### Planned Features
--- a/server/app.js
+++ b/server/app.js
@@ -48,6 +48,8 @@ var db = new sqlite3.Database('./messages.db', sqlite3.OPEN_READWRITE | sqlite3.
         sql += "pushsound TEXT, ";
         sql += "mailenable INTEGER DEFAULT 0, ";
         sql += "mailto TEXT, ";
+        sql += "telegram INTEGER DEFAULT 0, ";
+        sql += "telechat TEXT, ";
         sql += "ignore INTEGER DEFAULT 0 ); ";
     // initialise messages table
     sql += "CREATE TABLE IF NOT EXISTS messages ( ";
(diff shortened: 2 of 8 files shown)


s3m1s0n1c:
Looks good.. All works 👍

DanrwAU:
Fairly confident this works as expected, I tested a whole bunch last night with no issues. 
I also tweaked thee message format a little, if anyone can come up with anything better, feel free. 
Here's how it currently looks
image

davidmckenzie:
Awesome! Before we merge I'd like to have a think about #114 - perfect opportunity to solve that problem at the same time.

I also wanna have a quick look at the library's docos to see if we'll run into any issues with initiating `new telegram` on each message. If it's just setting up methods then should be fine, but if it's actually doing a pass of credentials at that point and retrieving another auth token or something, then could run into API limits later on.

DanrwAU:
`I also wanna have a quick look at the library's docos to see if we'll run into any issues with initiating new telegram on each message. If it's just setting up methods then should be fine, but if it's actually doing a pass of credentials at that point and retrieving another auth token or something, then could run into API limits later on.`

Good catch, the library i used doesn't seem to be the most active one, it was just the one the last fork had used, the other one seems to be more regularly developed but would have the same issue. 
https://github.com/yagop/node-telegram-bot-api

davidmckenzie:
Had a quick read of the current lib, it doesn't do anything fancy on its initial call, unless you have updates enabled, and they're off by default, so should be safe as is :)

It'd probably be more efficient and safer to set up the `t` variable up the top just after calling the library - unless there'd ever be a case where we'd want different API keys for different aliases. One key can be used for many channels, yeah?

DanrwAU:
Yep,

My understanding is the token is per bot, i can't see any use case for multiple bots so it should be safe to declare it once, the only consideration would be if it makes a call to validate - it may slow down message processing for messages without notifications

If we're going to go down that route the same probably applies to pushover, i can't ever seeing there being a need for multiple application keys (except to skirt their message limits which i wouldn't be writing in anyway) so it could be done at the top as well.

davidmckenzie:
You guys wanna review again? I've pulled the DB stuff out and added some logic to upgrade the DB schema when required.

It's controlled by the 'release' int in app.js, so all we need to do to push an update to schema is update that number and add the relevant bits into db.js. It will attempt to add missing columns as required. If more complex schema changes are required, we can probably add some logic in there to drop and recreate the tables.

Also moved the `t` definition up to the top - I haven't tested this bit yet, only tested the DB upgrade bits

marshyonline:
These changes look super good, Ill pull down tonight and submit a review 👍

s3m1s0n1c:
> Yep,
> 
> My understanding is the token is per bot, i can't see any use case for multiple bots so it should be safe to declare it once, the only consideration would be if it makes a call to validate - it may slow down message processing for messages without notifications
> 

Yeah the token is per bot..  I believe you can send messages to different channels/users from the same bot but only in the same instance.. you cant have 2 bots with the same name.

> If we're going to go down that route the same probably applies to pushover, i can't ever seeing there being a need for multiple application keys (except to skirt their message limits which i wouldn't be writing in anyway) so it could be done at the top as well.

Nah you shouldn't need multiple API keys as I don't we will hit the following criterias
- mess More then 30 messages per seconds
- mess More then 20 messages per minute 

They recommend not sending more then 1 message a second which may happen if the system spits out lots of pagers at the same time but they have the above limits

Maybe we can add tho to throw a error in log/console when 429 error is detected so we know the limit is being hit.

DanrwAU:
Tested Telegram with these changes, looks all good. 

DB Upgrade worked well on my production database as well (140k messages). 

Added a commit for a field i forgot to change the name of in the settings page :smile: 👍

s3m1s0n1c:
Yep Tested Telegram as well.. Works good..
DB Updated...

Can't review but good to go :)

DanrwAU:
> Date: (dateofmessage) Time: (Timeofmessage)
> Message: (Message)
> Brigade: (Alias) **Agency: (Agency)**

The agency should be shown in the title of the current template, not a fan of having the date and time, telegram has that anyway.

davidmckenzie:
Merging as is, we can make changes to message format later if needed :)

s3m1s0n1c:
Okays so Documentation for Telegram API has been done.

https://github.com/s3m1s0n1c/pagermon/wiki/Telegram

I plan on doing some more documentation on the rest of Pagermon when I get time.

Is this OK or do u want more detail/pics?

Thanks

marshyonline:
Love some good pics in guides

Could also make use of code blocks for things like examples :) 
Top work as always sonic

s3m1s0n1c:
Okays no worries.. Will add it in..

davidmckenzie:
Nice, thanks dude :) Have added you to collaborators if you wanted to edit direct on the repo's wiki

s3m1s0n1c:
> Nice, thanks dude :) Have added you to collaborators if you wanted to edit direct on the repo's wiki

No worries, Thanks

PR #123, opened 2018-11-07
Popovers for link collections

### LINKED ISSUE #47: Right click
Hey @davidmckenzie 
Would it be possible to enable the right click options to open links in new tabs on source capcode agency and alias on the index pages please? Atleast in Chrome these can only be opened by left clicking.

image

davidmckenzie:
Resolves #47

A bit of a weird one - basically converted all the dodgy-ass "links" (i.e. clickable td's with a redirect function) to open a popover with a collection of links.

At the moment the template for the popover is really barebones, but gives us something to work from for future enhancement, and knocks off that issue of not being able to right click the URLs, since they're all just standard anchors inside the popover now.

One downside is that it's an extra click when filtering, but I feel like the downside is worth it for future use cases. Think the advanced filtering type stuff etc.

Also in this is a bit of cleanup of the logic and dupe code of the various hiding things bits.

@jklnz may want to give it a crack too

### DIFF
--- a/server/public/templates/popover.html
+++ b/server/public/templates/popover.html
@@ -1,0 +1,19 @@
+<div class="popoverCont">
+    <div ng-if="popoverEl == 'source'">
+        <a class="btn" href="/?address={{message.source}}">Filter</a>
+    </div>
+    <div ng-if="popoverEl == 'address'">
+        <a class="btn" href="/?address={{message.address}}">Filter</a>
+        <a class="btn" ng-if="message.aliasMatch && user" href="/admin/aliases/{{message.aliasMatch}}">Edit Alias</a>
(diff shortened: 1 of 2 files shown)


DanrwAU:
This looks really good, updated my test box and tested it out, couldn't find any issues. 👍

DanrwAU:
Just a thought, is there a way to use this to open in a new tab rather than lose the page you are on? 

Sometimes i go back 10 or more pages to add aliases and it's a pain to end up back on page 1 between them and have to remember where i was up too  😭

davidmckenzie:
Well you can right click the links now to open in new tabs :D We could set default to open in new tab though for the create/edit links, probably makes sense. Would be nicer still to have them open a modal form for editing instead of needing another page load.

DanrwAU:
Right..... I'm dumb, never mind me 😜

davidmckenzie:
Changed to open the edit links in new tab, and merging :)

PR #126, opened 2018-11-10
Add time based duplicate filtering

### LINKED ISSUE #40: Duplicate messages ignored by time
Hey @davidmckenzie 
Just wondering if you could please add an option to ignore messages if it is within 20 (maybe have this user selectable?) seconds of the same message to the same address [as well as the current option to ignore if received in the last 10 messages].

davidmckenzie:
Resolves #40

Allows filtering of duplicate messages based on a time limit (e.g. past 60 seconds) as well as or instead of the current 'last x messages' filtering.

### DIFF
--- a/server/config/default.json
+++ b/server/config/default.json
@@ -12,6 +12,7 @@
     "defaultLimit": 20,
     "duplicateFiltering": true,
     "duplicateLimit": 10,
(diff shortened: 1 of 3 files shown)


davidmckenzie:
Whoops, good catch, thanks :D

PR #133, opened 2018-11-16
API Sec Mode - fix new messages not showing capcode

marshyonline:
This Fixes capcodes not showing when logged in in api sec mode

### DIFF
--- a/server/views/index.ejs
+++ b/server/views/index.ejs
@@ -265,9 +265,8 @@ var app = angular.module('app', ['ngRoute', 'ngResource', 'ngCookies', 'angular-
     <% } %>	
   <% } else { %>
     <% if (login) { %>
-      var socketMode = socket
-      socket.open();
-      adminSocket.close()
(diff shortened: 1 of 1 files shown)


DanrwAU:
Good Work, I've been meaning to look into this!  

I'll  pull it down shortly and test it out 👍

DanrwAU:
This one's not working for me, index is no longer being updated with new messages, the messages are showing in the websocket from developers tools 😕

davidmckenzie:
This whole chunk of stuff is confusing :D 

@DanrwAU does it work when you have apisecurity disabled, but hidecapcode true? Should be the same code I think

DanrwAU:
Definetly works with HideCapcode on - doesnt work in secure mode. 

The messages asre definetly showing in the socket tho 😦

marshyonline:
Ok, that should be easy then

marshyonline:
Ive broken git 😆 
Leave this one with me

Issue #130, opened 2018-11-14
Refactor delivery methods into plugin system


davidmckenzie:
Have some kind of extensible plugin system that allows users to pick and choose which notification methods they'd like to use. Need to consider DB structure - pop all the notification methods out into a JSON string?

Possibly use something like https://github.com/c9/architect

davidmckenzie:
FYI @marshyonline @DanrwAU @s3m1s0n1c @all3kcis

I've started on this in the 'plugins' branch here: https://github.com/davidmckenzie/pagermon/tree/plugins

Still a lot to do but just as a heads up as the structure of the plugins is starting to take shape. The configuration definitions for each plugin will be stored in the plugins dir, but configuration itself will remain in the main config.json, just in a bit of a different format.

Global configuration for each plugin is pretty much done - the definitions have been copied into the relevant files and the config can be edited in UI no dramas. It's all ng-repeat so in theory infinitely scalable. Conversion of old config to new is also sorted, handled in the db upgrade thingo.

Hooks have been added into the message API route and is triggering the plugins all good, they just don't do anything yet. The plugins can be run either before or after message processing - I'm not sure of the use for the 'before' plugins, but I'm sure there's something neat that can be done there. I've tried to keep it forward-compatible as much as possible, so there's an 'event' variable that's just 'message' for everything at the moment, but thinking ahead we may want to add a plugin to extend something else, like the capcodeRefresh function or maybe something to intercept page views.

Next up for me will be converting the config on the alias level - I'll be doing this as a single text field on the capcodes table that stored a json variable. Only real way to keep it extensible, but should (hopefully) get the job done nicely. The DB upgrade scripts for this bit will be a little scary, as they'll need to do a recreate of the capcodes table - lots of places that can go wrong.

Once that's done it'll just be a matter of ripping out all of the code in the message API route into the individual modules, and a buttload of testing.

Now that I've typed it out it sounds like a lot more work than I thought it was 😁

davidmckenzie:
Ok so this is kinda functionally done now. That was one of those all-day marathons where I forget to eat so it is almost certainly filled with bugs and stuff.

I've tested basically nothing of this so far, and documented nothing, but this could be considered ready for testing by the enormously brave.

**If you've got big enough balls to try this - please make sure you triple back up your messages.db file. It will probably blow up or something.**

[...]

PR #134, opened 2018-11-17
Ignore not matching capcodes

all3kcis:
Use to ignore all messages where address not matching with the patern (javascript regex)

### DIFF
--- a/server/config/default.json
+++ b/server/config/default.json
@@ -13,6 +13,7 @@
     "duplicateFiltering": true,
     "duplicateLimit": 10,
     "duplicateTime": 60,
+    "ignoreNotMatching": "",
     "rotationEnabled": true,
     "rotateDays": 7,
     "rotateKeep": 4,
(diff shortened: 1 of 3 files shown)


davidmckenzie:
Thanks for the PR! Not sure about this one though - what's it trying to solve that isn't solved by existing aliases with the filter checkbox selected?

You can already create aliases that are ignored with wildcard matching, so it seems like it would only really fit a very specific use-case where complex regex is required for the filtering, unless I'm missing something?

In the example given in the help text, you could just as easily create an alias with the address '00%' to filter out those messages

all3kcis:
Yes, I need to ignore all message that match with /^(0|1)86, it is not possible with actual alias system

DanrwAU:
I had to do something similar for an instance I setup for someone.

I ended up doing it on the pocsag receiver (in reader.js).

davidmckenzie:
Going to close off this PR as I've implemented as a plugin in #130 / #135 - hit me up if this isn't what you had in mind and we can tweak the plugin :)

PR #135, opened 2018-11-17
Add exec shell command on selected alias

all3kcis:
Like Pushover or email alert, that's add the possibility to exec shell command for selected aliases.

I developt it because i'm a firefighter and I use Pagermon with this feature to turn on the light on night hours if my pager ring.

here is an example to use :
On my personnal alias :
capture

And in lights.py ( for example )
```python
from datetime import time, datetime

#### VARS
now = datetime.now()
# Night hours
start_time = time(22, 00)
stop_time = time(9, 00)
# Light duration
duration = 60 * 8	


### CODE

if ( now.time() >= start_time or now.time() <= stop_time or True):
	# Exec script
	## your script here
else:
	print 'Time not match'
```

### DIFF
--- a/server/config/default.json
+++ b/server/config/default.json
@@ -56,6 +56,9 @@
     "teleenable": false,
     "teleAPIKEY": ""
   },
+  "shellCmd":{
+    "shellCmdEnable": false
+  },
   "STMP":{
     "MailEnable": false,
     "MailFrom": "",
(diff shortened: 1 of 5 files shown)


marshyonline:
One the merge conflicts are resolved there is a +1 from me

all3kcis:
Oh yes...
Maybe I can just use enabled / disabled option on aliases and exec file stored in folder pagermon/local-scripts ?
For example : Alias id 52 with "ShellCmd" option enabled, when a message address match, we looking for "pagermon/local-scripts/alias-52.sh" and exec it if exist.

Or #130 is an another solution yes

davidmckenzie:
I'm gonna put some work into #130 today, will see how far I get there and let ya know :)

all3kcis:
Please close and see #154

PR #145, opened 2018-11-27
Modify minimum message length for POCSAG1200

DanrwAU:


DanrwAU:
Fixes https://github.com/davidmckenzie/pagermon/issues/144

davidmckenzie:
Big change, we might have to increment major release number 😁

PR #139, opened 2018-11-18
Plugin framework

davidmckenzie:
Solves #130 

**No merge - still WIP, PR for visibility only**

PR #148, opened 2018-11-30
Sort messages by timestamp

### LINKED ISSUE #146: Sort messages by date
I populated the database manually with  "old" messages. They are sorted by `id`.
See #41

pyrog:
Fix #146


davidmckenzie:
I'm gonna incorporate this into #139 to save the hassle of rebasing/merging. One thing to note though - have you noticed any issues with performance? I'd have to run some explain query's, but just thinking whether indexes would be impacted.

PR #149, opened 2018-12-01
Fix #147

### LINKED ISSUE #147: Duplicated column Aliase
duplicated column aliases

This happened for **non-logged in users** only.

pyrog:


davidmckenzie:
Per comments in #147, this is already fixed in #139. Thanks for the PR though :)

PR #153, opened 2018-12-06
Discord Plugin fixes

marshyonline:
Util is required here
Sill trying to work out why its bouncing back from the Discord API
```[pmon]  2018-12-06 23:12:08  info : Discord: Message Sent
[pmon]  2018-12-06 23:12:35  error : Invalid Form Body
embeds[0].author.url: Not a well formed URL.
```
Getting the above error with the hostname filled or not filled

### DIFF
--- a/server/plugins/Discord.js
+++ b/server/plugins/Discord.js
@@ -1,6 +1,7 @@
 var discord = require('discord.js');
 var toHex = require('colornames');
 var logger = require('../log');
+var util = require('util');
 
 function run(trigger, scope, data, config, callback) {
     var dConf = data.pluginconf.Discord;


DanrwAU:
Strange... I thought I fixed that already! 

That error look familiar but I can't remember what the cause was, I'll take a look tomorrow if you haven't figured it out :)

marshyonline:
looks like its ignoring the hostname from process.json and returning my servers hostname (thats not a valid url) and discord doesn't like that

marshyonline:
Ok - PM2 was just being a bitch and I need to tell it when the process.json fill was again for it to reload the evn's from it. 

The above change is all that should be required now

davidmckenzie:
Merging, thanks dude!

PR #171, opened 2019-01-04
WIP Msg clumping

davidmckenzie:
No mergey - still WIP, PR for visibility

### DIFF
--- a/server/public/templates/popover.html
+++ b/server/public/templates/popover.html
@@ -1,19 +1,38 @@
 <div class="popoverCont">
-    <div ng-if="popoverEl == 'source'">
-        <a class="btn" href="/?address={{message.source}}">Filter</a>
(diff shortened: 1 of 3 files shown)


DanrwAU:
This lags my page loads considerably as expected, needs a settings toggle to turn it off methinks!

davidmckenzie:
Thanks dude :) Will have to think on this one, there's gotta be a better way to do it

davidmckenzie:
Yo @DanrwAU I just noticed we're missing an index after the change to sort by timestamp - will add this in a separate PR, but wanna test with this index added and see if it makes a difference?

```sql
CREATE INDEX IF NOT EXISTS `msg_timestamp` ON `messages` (`timestamp` DESC, `alias_id`);
```

Issue #76, opened 2018-06-20
Queuing pager messages on client machine when server is not contactable.


s3m1s0n1c:
Thank you for this awesome program,

Its a shame i've only just discovered it as i've been using another program but like this one as u can do filtering and it deals with duplicates more cleanly.

Does Pagermon have some sort of buffer/queuing system if the server is not able to be contacted?

Eg If Internet drops from client and it receives a pager message then it drops the pager messages cause it cant contact the server.

Solution is to have some sort of file it send messages too and reader.sh reads off it and deletes the row when finished.

Moving to SQL DB would solve this also as u can sync DB's which is what I do atm.

Food for thought.. Maybe this can be a feature one day.

davidmckenzie:
Could add some looping in the catch on line 114, if it fails then send it off to a setTimeout driven function that retries every x seconds. Might lead to memory leakage if the server is down for a while though.

davidmckenzie:
Closing as solved in #173, I'd like to revisit this at some point though

PR #173, opened 2019-01-05
Flex frag + retries

### LINKED ISSUE #108: FLEX multipart Messages
Hi All,
    Im trying to contribute to this project but as I'm from a python programming no JS I'm having some issues joining the messages together

Multimon now supports fragmented messages by using a letter being K (complete page) F (fragmented message) and C (continued) . This can be found in the FLEX string here being represented by the *

`1600/2/ * /A`  

what my aim was in using reader.js to add these stings together and then upload.

an example sting as follows
`2018-10-22 02:06:45 1600/2/F/A [001234567] ALN TH`
`2018-10-22 02:06:45 1600/2/C/A [001234567] ALN IS IS A TEST PAGE`
`2018-10-22 02:07:45 1600/2/K/A [001234567] ALN THIS IS A SECOND TEST PAGE`

i can see in reader.js regex is used to match the word FLEX. Is it possible to add another regex to find the find the `F` and store the message and capcode. Then if the next  page contains the `C` compare the capcodes and if they match join up the messages and upload. Then the other strings can be cleared.

Hope this makes sense.

happy to test and supply more examples or can save audio if required

Shaggs

davidmckenzie:
Resolves #108 and #76 

/cc @Shaggs and @112bzo as FLEX users

This moves the FLEX fragmented messages into an object with the address as key - fragmented messages that arrive with other messages in between should now properly join. The only real edge case I can think of is if in poor reception and you receive an F message, but never its corresponding C message, then you might one day receive a randomly joined message. Would be rare enough not to worry about, IMO.

The other change in this is the retries that @s3m1s0n1c requested - messages will retry sending on failure up to 10 times, with the time between retries increasing exponentially (`Math.pow(2, retries)`). This isn't the best way to handle this, especially for machines that will be offline for some time. The alternative would be to use a proper queuing system like Redis, but that would greatly increase the barrier to entry and reduce compatibility/portability. Probably something to look at if there was ever a demand for a "production ready" pagermon.

### DIFF
--- a/client/reader.js
+++ b/client/reader.js
@@ -63,16 +63,16 @@ rl.on('line', (line) => {
   // TODO: pad address with zeros for better address matching
 //  if (line.indexOf('POCSAG512: Address:') > -1) {	
   if (/^POCSAG(\d+): Address: /.test(line) ) {
-  	address = line.match(/POCSAG(\d+): Address:(.*?)Function/)[2].trim();
+    address = line.match(/POCSAG(\d+): Address:(.*?)Function/)[2].trim();
     if (line.indexOf('Alpha:') > -1) {
-    	message = line.match(/Alpha:(.*?)$/)[1].trim();
-    	trimMessage = message.replace(/<[A-Za-z]{3}>/g,'').replace(/Ä/g,'[').replace(/Ü/g,']');
+      message = line.match(/Alpha:(.*?)$/)[1].trim();
+      trimMessage = message.replace(/<[A-Za-z]{3}>/g,'').replace(/Ä/g,'[').replace(/Ü/g,']');
     } else if (line.indexOf('Numeric:') > -1) {
       message = line.match(/Numeric:(.*?)$/)[1].trim();
       trimMessage = message.replace(/<[A-Za-z]{3}>/g,'').replace(/Ä/g,'[').replace(/Ü/g,']');
     } else {
-    	message = false;
-    	trimMessage = '';
+      message = false;
+      trimMessage = '';
     }
   } else if (line.indexOf('FLEX: ') > -1) {
     address = line.match(/FLEX:.*?\[(\d*?)\] /)[1].trim();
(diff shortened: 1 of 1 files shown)


client/reader.js
```diff
@@ -106,37 +107,50 @@ rl.on('line', (line) => {
...
 }).on('close', () => {
   console.log('Input died!');
 });
 
+var sendPage = function(message,retries) {
+  var options = {
+    method: 'POST',
+    uri: uri,
+    headers: {
+      'X-Requested-With': 'XMLHttpRequest',
```
marshyonline: Can we add 
`'User-Agent': "reader.js",

marshyonline: This will make filtering out API requests from decoders a simple task

davidmckenzie: Makes sense... will that mess with CloudFlare at all? Haven't checked what user-agent it sends at the moment

davidmckenzie: Looks like it's only an issue if you manually enable UA filtering so should be all good

davidmckenzie: Added as "PagerMon reader.js" :)

Shaggs:
@davidmckenzie I'll test tonight.  In regards to reception even doing this in python or c++ has the same potential. Nothing is perfect in life minor issues like that can't be helped. 

Our old FELX system didn't have this issue but it's all about streamlining and getting pages out faster. 

Thanks for all your time and effort

marshyonline:
Had a thought re: the message retries.

We could store the messages locally in a sqlite db if unable to send the first time - set a timer and ping a API endpoint so see if the server is alive then resend any messages pending being delivered

davidmckenzie:
Thanks guys! Gonna merge this as is as got a few things in the queue, will probably revisit later for something more permanent

s3m1s0n1c:
@davidmckenzie
Thanks will try soon

PR #222, opened 2019-02-06
Plugin - MessageRepeat

marshyonline:
# Description

Allows the server to forward all incomming messages to another Pagermon Server / API

## Type of change

# How Has This Been Tested?

Ran local test's and on my production servers.
Code also passed through eslint

# Checklist:

### DIFF
--- a/server/plugins/MessageRepeat.js
+++ b/server/plugins/MessageRepeat.js
@@ -12,6 +12,7 @@ function run (trigger, scope, data, config, callback) {
   if (config.repeatURI) {
     var uri = config.repeatURI
     var apikey = config.repeatAPIKEY
+    var do_not_forward = false;
 
     if (config.repeatUUID == 0 || !config.repeatUUID) {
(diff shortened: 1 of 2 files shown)


marshyonline:
Only issue I can see with this is when you point it at the same server it ends up in an infinite loop
Changing it's Scope to afterMessage would only send messages that have matched if running in PDW mode?

davidmckenzie:
Could add an extra field in the message data. If that field exists and == some kind of unique identifier for the current box, then ignore. This could all be handled in the same plugin since the whole `req.body` object is passed through.

Plugin could generate a random UUID and store it in the pluginConf, then send that as `relayID` or something.

marshyonline:
Ohhhh that's a good idea.
BeforeMessage can make the server ignore a message. 

I'll have a poke tomorow

DanrwAU:
Did a quick install and test, looks good. 👍

marshyonline:
Couldnt think of a nice way to auto set the UUID
so we force the user to user a UUID on config 
Servers will now no longer loop there own messages :D

PR #174, opened 2019-01-05
Support Multiple Databases

DanrwAU:
# Description

Complete re-write of database driver to remove native SQLite 3 client and replace with Knex query builder.

Add support for MySQL with possiblity of Amazon RDS and Postgres to be implemented later

Fixes #10 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

**Need to Enter Tests Completed**

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated changelog.md with changes made

PR #243, opened 2019-03-27
Better Docker support for server

nicko170:
# Description

This actually makes docker server usable, keeping configuration persistant across container reloads by migrating to an entry point script that creates files that do not exist and linking to required folders.

Fixes # (issue) N/A

## Type of change

# How Has This Been Tested?

- [ ] docker build .
- [ ] confirmed container started and app running
- [ ] added to docker hub and deployed in production :-P - https://hub.docker.com/r/nick170/pagermon 

# Checklist:


davidmckenzie:
Neat! I'll let the other @pagermon/collaborators review as I don't have a docker env kicking about.

I wonder if there's any value wrapping in pm2 as well, or is that not really needed with dockers magical healthchecky stuff?

Only other thing you might want to add is the env variables https://github.com/pagermon/pagermon/blob/a81c7b900c0f2eceb4c07c958b23909f252df028/server/process-default.json#L20-L25

nicko170:
Augh -- Yeah, I should add them in. I'll add another commit in a few minutes to support environment variables properly.

I think it *should* be wrapped in PM2 -- but docker kind of does take care of the load-balancy/ingress routing side -- and will moreso once I complete the next bit that I want to fix.

I was going to refactor for Redis / node-busmq and maybe mongodb in the end (for real-time load-balancy stuff) next, because then my production docker cluster can just keep this running happily without needing to care about losing data and can have automatic clustering (I can only have a single instance running at the moment - but I want scale-out) -- but I needed docker actually working first.

The way I see it - post a message to the API, it drops it into messagequeue, and a process from there then saves to DB on a single node - but each node (and there for each potential web socket) will see the new message come in on the queue and have it available for websockets.

marshyonline:
Ah and can you please also update the Change Log for the changes

davidmckenzie:
> I was going to refactor for Redis / node-busmq and maybe mongodb in the end (for real-time load-balancy stuff) next, because then my production docker cluster can just keep this running happily without needing to care about losing data and can have automatic clustering (I can only have a single instance running at the moment - but I want scale-out) -- but I needed docker actually working first.

Man that's the dream. I really like the idea of using redis for the message processing, has been something I've wanted to give a crack for a while now. The upcoming changes in #174 definitely get us closer to being able to run load balanced too, just need the mq aspect and some rudimentary session state sharing really.

Still want to support mongo at some point too, would give us a lot more flexibility with db structure stuff, but at that point it might even need to evolve into a separate fork. PagerMongo? 😁

nicko170:
@marshyonline  -- I've added some more changes for docker. Namely, what you have requested, but moved to docker-compose so when redis / mongo is required, it can just be added there in future. That needs to remain in base directory -- but docker file has been moved into server per request.

Environment variables are now passed along, too :-)

Nicks-MacBook-Pro:pagermon nickp$ docker-compose up -d
Starting pagermon ... done
Nicks-MacBook-Pro:pagermon nickp$ docker-compose exec pagermon sh
/app # cat process-default.json
{
  "name"             : "pagermon",
  "cwd"              : "/app",
  "script"           : "app.js",
  "node_args"        : ["--harmony"],
  "error_file"       : "logs/node-app.stderr.log",
  "out_file"         : "logs/node-app.stdout.log",
  "pid_file"         : "pids/node-geo-api.pid",
  "instances"        : 1, //or 0 => 'max'
  "min_uptime"       : "200s", // 200 seconds, defaults to 1000
  "max_restarts"     : 10, // defaults to 15
  "max_memory_restart": "100M", // 1 megabytes, e.g.: "2G", "10M", "100K", 1024 the default unit is byte.
  "cron_restart"     : "1 0 * * *",
  "watch"            : false,
  "ignore_watch"      : ["[\/\\]\./", "node_modules", "config", "logs"],
  "merge_logs"       : true,
  "exec_interpreter" : "node",
  "autorestart"      : true, // enable/disable automatic restart when an app crashes or exits
  "vizion"           : false, // enable/disable vizion features (versioning control)
  "env": {
    "NODE_ENV": "production",
    "HOSTNAME": "localhost",
    "USE_COOKIE_HOST": false,
    "APP_NAME": "pagermon"
  }
}
/app #

Should probably incriment the version number in the app, stand by :-)

server/app.js
```diff
@@ -1,5 +1,5 @@
-var version = "0.2.3-beta";
-var release = 20181118;
+var version = "0.2.4-beta";
+var release = 20190327;
```
DanrwAU: The release var only needs to change if the DB structure is changed. So this one isn't required (in fact it breaks some of my logic on #174 lol.

I think we're pushing for the next release to be 0.3.0 we usually merge to the version branch then do a pull to master for the release.

DanrwAU:
@pagermon/collaborators do we want to include this into the 0.3.0 branch first before pushing to master? I don't think a dockerfile warrants a full release (just my opinion 😄 )
I'd prefer this commit https://github.com/pagermon/pagermon/pull/243/commits/bddb0a52e3ef1db787d3687ce7c2cf722feddd3a  get's rolled back and we put it into 0.3.0.

One other question i had, baring in mind I'm not really familiar with docker, but why are you removing then recreating the process-default.json file?

server/app.js
nicko170: LOL. No problems. How do you want me to add this to the change log? Just under 0.2.3 or?

DanrwAU:
> One other question i had, baring in mind I'm not really familiar with docker, but why are you removing then recreating the process-default.json file?

Ah i see now.... 

Is it possible to use sed or similar to modify the existing file, just so if that file ever needs to be modified from our end we don't need to remember to change in 2 locations? 😀

nicko170:
> One other question i had, baring in mind I'm not really familiar with docker, but why are you removing then recreating the process-default.json file?

There are two ways to do this -- using a bunch of sed to manipulate the file (its how my spamassassin container does it) -- but that gets really messy, really quickly. It's also 10PM here and I am tired so sed/regex is somewhat missing from my brain right now :-P

OR we can just replace the whole file, every time the container restarts. As its built on env variables which are in the docker-compose.yml file, it cannot be broken / forgotten so I don't really see an issue. It would be the same either way. But I opted for this path as the cwd value (/home/ubuntu/workspace/server) was wrong for the container was wrong and required changing.

nicko170:
> Is it possible to use sed or similar to modify the existing file, just so if that file ever needs to be modified from our end we don't need to remember to change in 2 locations? 😀

HAHAHAHA. Sure. I'll change to sed and push a new change through.

Want me to also rollback bddb0a5 while I am here?

DanrwAU:
> > Is it possible to use sed or similar to modify the existing file, just so if that file ever needs to be modified from our end we don't need to remember to change in 2 locations? 😀
> 
> HAHAHAHA. Sure. I'll change to sed and push a new change through.
> 
> Want me to also rollback [bddb0a5](https://github.com/pagermon/pagermon/commit/bddb0a52e3ef1db787d3687ce7c2cf722feddd3a) while I am here?

I agree SED is painful, in fact i never got it working last time i tried, but i think that's the best way to only change what you need to and not double up on code/files. 
The version number and release can be rolled back and then we might try and rebase to the 0.3.0 branch and see what happens 😕

nicko170:
Should be good now for SED instead of cat -- it breaks the tabspaces in the file, but semantics, right?

Nicks-MacBook-Pro:pagermon nickp$ docker-compose up -d && docker-compose exec pagermon cat process-default.json && docker-compose down
Creating network "pagermon_default" with the default driver
Creating pagermon ... done
{
  "name"             : "pagermon",
   "cwd" : "/app",
  "script"           : "app.js",
  "node_args"        : ["--harmony"],
  "error_file"       : "logs/node-app.stderr.log",
  "out_file"         : "logs/node-app.stdout.log",
  "pid_file"         : "pids/node-geo-api.pid",
  "instances"        : 1, //or 0 => 'max'
  "min_uptime"       : "200s", // 200 seconds, defaults to 1000
  "max_restarts"     : 10, // defaults to 15
  "max_memory_restart": "100M", // 1 megabytes, e.g.: "2G", "10M", "100K", 1024 the default unit is byte.
  "cron_restart"     : "1 0 * * *",
  "watch"            : false,
  "ignore_watch"      : ["[\\/\\\\]\\./", "node_modules", "config", "logs"],
  "merge_logs"       : true,
  "exec_interpreter" : "node",
  "autorestart"      : true, // enable/disable automatic restart when an app crashes or exits
  "vizion"           : false, // enable/disable vizion features (versioning control)
  "env": {
   "NODE_ENV" : "production",
   "HOSTNAME" : "localhost",
   "USE_COOKIE_HOST" : false,
   "APP_NAME" : "pagermon"
  }
}
Stopping pagermon ... done
Removing pagermon ... done
Removing network pagermon_default

marshyonline:
I will take a look in the morning 
Should be able to fix the tab issue 

Great work dude 👌

nicko170:
Yeah -- about that... HEH.. Maybe hold off :-)

Just about to push a new commit that moves to pm2, and will rebase my changes into knex/v0.3.0 branch and add support for setting default mysql configuration through env variables, and updating docker-compose.yml to spin up mysql server -- so docker-compose up will give a full stack environment that just works.

Stand by. Once this is done, it should just continue working when 0.3.0 becomes master instead of breaking then

DanrwAU:
Should be right to rebase to 0.3.0 I'm keeping the knex build up.to date as required.
Good work!

davidmckenzie:
This is looking good, thanks dude! 👌

nicko170:
Annnnnnd we are good now :-)

PR #244, opened 2019-03-27
Proper docker support

nicko170:
See: #243 -- This superseeds #243 which should now be rejected / closed.

I *think* I have un-broken everything that I broke, it's compiling, working, and was rebased from v0.3.0 so should remain stable :-)

Change log is also fixed.


marshyonline:
I might let the other @pagermon/collaborators look at this one - your branch seem to be wanting to pull lots more then what it should be.
you should be ok to use your other PR - all that would be needed is to resolve the CHANGELOG conflict

DanrwAU:
I'm not sure what GIT is doing with all those commits, tried to clean it up from the web interface but it's not having a bar of it. You also seem to have had some issues with the merge commit in the changelog, somethings gone very wrong there 😉 

I'd like to clean that up first as recommiting the same changes will look messy as. 
Project for this afternoon 😄

nicko170:
Yeah -- I am fixing it quickly. I'll kill this one off and revert the changelog in 243, and that should do it ;-)

I did say I was tired last night -- clearly should not be allowed near a computer whilst tired heh

PR #241, opened 2019-03-24
High Resolution Mobile Fix

DanrwAU:
# Description

Fixes issues with high-resolution mobile devices having the index page squished by long values in the alias column. 

Pre-Fix
image

Post-Fix
image

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
Test homepage loading pre and post change on a variety of resolutions. 

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated changelog.md with changes made

PR #253, opened 2019-04-09
Fix High Resolution Display Properly

DanrwAU:
# Description

Fixes Mobile properly - leading on from #241 and fixing properly

## Type of change

# How Has This Been Tested?

Tested Display on Galaxy S10 which was not affected by changes in #241 

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * Convince Winston to follow the laws of Physics #251
 * More complete docker support #243, #249
-* Fix display issues on large resolution mobile devices #241
+* Fix display issues on large resolution mobile devices #241, #253
 * Update old repository URL in package.json #240
 * Add ability to select the icon size on the homepage #239
 * Fix 404 error when trailing slash added to hostname in client config #223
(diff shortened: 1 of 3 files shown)


DanrwAU:
@pagermon/collaborators This continues on from #241 and seems to properly fix it  on my devices 👍

davidmckenzie:
Hmm, this might break resized windows, like if you had a browser window open on desktop but only half screen width. Might need to duplicate that so it runs on both max-device-width and max-width.

Dunno how easy it is to get on the s10, but are you able to find what it's reporting for the device-width and the actual viewport width?

DanrwAU:
Device width is 412px max width is 1440px. 

Would setting the device orientation to filter only devices in portrait help or should we add NoTiny
Back in as well?

davidmckenzie:
I think it's more that max-width *should* be doing what you need, if I understand it correctly. Since max-width is the width of the current browser window, and max-device-width is the max width reported by the device itself (like the screens actual resolution), in theory max-width would always be the more accurate number. Unless something funky happening with upscaling that makes it not do stuff right.

Upscaling fuckery is the only reason I can think of that max-width would be higher than max-device-width

DanrwAU:
@davidmckenzie The last 2 commits introduce a noSmall class with the old parameters which fixes resizing 😄

PR #283, opened 2019-05-08
Adding POCSAG "Tone Only" support !

all3kcis:
Added support for POCSAG "Tone only" messages (Empty message)

# User side : 
A first parameter in the options allows you to choose if you want to record "Tone only" messages, three values are possible: (Always, For selected aliases, Never), Never is by default.  
settings

If the choice is "For selected aliases" it becomes possible on the alias page to choose if you want to record the tone only for this alias.  
alias

Tone only as displayed like "Tone only" :)  
show

# Developer side :  
data message is now tagged with "data.isToneOnly" parameter, it can be true or false

# To be discussed
"Tone only" is stored as another message without distinction, maybe it might be interesting to add an "isToneOnly" column in the database?

# Wiki updates
I suggest to create a section "Messages"  and adding
```
# Messages
## Store "Tone Only" messages  
I can be possible to store "Tone only" message, just set option to "Always" or "For Selected aliases".
If you choose "For Selected aliases" you must check "Store Tone only"  for the desired alias.
> **Warning**, other options or plugins may choose not to store the message.
>    
> **If despite the "Always" option you do not receive the "Tone only" messages, refer to the "Client Installation" section**
```
Clients section, just add  
```
**Receving "Tone only" message.**  
To receive "Tone only" messages, make sure you do not use the "-e" setting in multimon-ng.
```
I still have to check if the "-f alpha" parameter should be removed as well, but I think so.

### DIFF
--- a/client/config/default.json
+++ b/client/config/default.json
@@ -3,5 +3,6 @@
   "hostname": "http://127.0.0.1:3000",
   "identifier": "TEST",
   "sendFunctionCode": false,
-  "useTimestamp": true
+  "useTimestamp": true,
+  "enableToneOnly": true
 }
--- a/client/reader.js
+++ b/client/reader.js
@@ -97,8 +97,8 @@ rl.on('line', (line) => {
       message = line.match(/Numeric:(.*?)$/)[1].trim();
       trimMessage = message.replace(/<[A-Za-z]{3}>/g,'').replace(/Ä/g,'[').replace(/Ü/g,']');
     } else {
-      message = false;
-      trimMessage = '';
+      message = "TONE ONLY";
+      trimMessage = message;
     }
   } else if (line.match(/FLEX[:|]/)) {
     address = line.match(/FLEX[:|] ?.*?[\[|](\d*?)[\]| ]/)[1].trim();
--- a/client/reader.js
+++ b/client/reader.js
@@ -31,6 +31,7 @@ var apikey = nconf.get('apikey');
 var identifier = nconf.get('identifier');
 var sendFunctionCode = nconf.get('sendFunctionCode') || false;
 var useTimestamp = nconf.get('useTimestamp') || true;
+var enableToneOnly = nconf.get('enableToneOnly') || true;
 
 //Check if hostname is in a valid format - currently only removes trailing slash - possibly expand to validate the whole URI? 
 if(hostname.substr(-1) === '/') {
@@ -97,8 +98,13 @@ rl.on('line', (line) => {
       message = line.match(/Numeric:(.*?)$/)[1].trim();
       trimMessage = message.replace(/<[A-Za-z]{3}>/g,'').replace(/Ä/g,'[').replace(/Ü/g,']');
     } else {
-      message = "TONE ONLY";
-      trimMessage = message;
+      if(enableToneOnly){
+        message = "TONE ONLY";
+        trimMessage = message;
+      }else{
+        message = false;
+        trimMessage = '';
+      }
     }
   } else if (line.match(/FLEX[:|]/)) {
     address = line.match(/FLEX[:|] ?.*?[\[|](\d*?)[\]| ]/)[1].trim();
--- a/client/samples/testToneOnly.sh
+++ b/client/samples/testToneOnly.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+cd .. && echo "POCSAG512: Address: 1000000  Function: 0 " | node ./reader.js
--- a/server/config/default.json
+++ b/server/config/default.json
@@ -16,6 +16,7 @@
     "duplicateFiltering": true,
     "duplicateLimit": 10,
     "duplicateTime": 60,
+    "storeToneOnly": "never",
     "rotationEnabled": true,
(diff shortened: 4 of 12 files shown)


davidmckenzie:
Imo the decision to store shouldn't be on the server side - should be on the client side whether to send or not. It's a lot of added complexity for a decision the client side can make rather easily, and for a feature that only very few users will adopt.

If filtering on specific capcodes is required then I would suggest using the plugin framework for that rather than introducing DB schema changes

DanrwAU:
I'd prefer this one as a client side as well, there isn't any need for the server to do any special processing if the client side figures it out first and modifies. 
The specific capcode stuff could be included with the client side capcode list requested the other day on discord?

all3kcis:
Hi, What do you think of my last proposal?

davidmckenzie:
Why does it need a plugin? If you're modifying the client side to optionally parse tone only then there shouldn't be any need for any logic on the server side at all. All the server side sees is a normal message that happens to have "TONE ONLY" as the content of the message. It seems redundant to have the decision take place in two places.

all3kcis:
The plugin is used to define if the message will be stored or not, but if the plugin is enabled the message will be processed. (by other plugins if necessary) 
This will make it possible to interact with this type of message. 
I find it interesting to be able to perform actions when receiving a "tone only" without wanting to store it in a database.

If you are against the idea of this plugin I will only submit the modification of the client part.

davidmckenzie:
Just thinking over it again yeah I can see the value in the plugin mostly for enabling tone only messages on specific aliases only. Like over here I know the NSW guys would want to store tone messages for FRNSW, but would want to ignore for RFS.

I'd default the plugin to enabled and default to store for all - that way the plugin is completely seamless for most users who want it to just behave like every other message.

all3kcis:
> I can see the value in the plugin mostly for enabling tone only messages on specific aliases only.  

It's kind of the use I need :)

Can you summarize for me what is your vision of how this new feature "Tone only" works?

I wonder how to ensure that a "Tone Only" message is not processed like an ordinary message by plugins.
For example, I think that by default a "Tone only" message should not be sent by Pushover  without  user's expressive agreement.

Maybe with something  like that 
```javascript
data.isToneOnly = false;
if ( data.message = "TONE ONLY"  ){
  data.isToneOnly = true;
}
// or
data.messageType = "toneonly"
```
But this will require an update of all plugins...

davidmckenzie:
Hmm I'm not sure you'd want to assume any different behaviour on the plugins. If you wanted to pass that through to other plugins it would need an extra `before` plugin to identify that it's tone only. The `before` plugins can't do alias based rules since it's before any alias matching, so would need to be an additional plugin. The `after` plugins can still modify the pluginData variables, but there's no ordering of them, so no way of guaranteeing that one will run before the others.

Probably safer to duplicate the logic in any plugins you want to modify to react differently to tone only messages. The `if ( data.message = "TONE ONLY"  )` thing being in all relevant plugins isn't that much tech debt, and probably wouldn't need to be in all of them.

all3kcis:
Can you tell me how you want me to implement this feature ?

all3kcis:
@davidmckenzie

davidmckenzie:
Not sure what you mean? If you're referring to this bit:

> Probably safer to duplicate the logic in any plugins you want to modify to react differently to tone only messages. The if ( data.message = "TONE ONLY" ) thing being in all relevant plugins isn't that much tech debt, and probably wouldn't need to be in all of them.

Then that's just a matter of modifying the pushover plugin (for example) to have an extra config item for yes/no on tone only, right?

Unless I'm missing something? :)

all3kcis:
# Client side: 
By default "Tone only" messages are accepted and sent to the server.
It is possible to disable them from the client conf file. ( set to false )

# Server side :
Message settings: 2 new settings
- "Process tone only messages," "Never," "Only for selected alias," "Always,"
If "Never", which is the default value, the server does no processing at all, so the impact of these messages is very low
- "Store tone only messages," "Never," "Only for selected alias," "Always,"

It is therefore possible to process "ToneOnly" with compatible plugins, but not to store the message.
image

# Plugins
Plugins must declare themselves compatible (by adding, "acceptToneOnly": true in the json config file)
So no surprises.
image



# Aliases
one new parameter  
It is dependent on the setting made in the general configuration.
- "Store tone only" if the box is checked the message will be stored and processed by plugins.

# Aliases - Plugins part
"Process tone only", the message will be processed by the plugin
It allows to choose to process tone only message only on selected comaptible plugins
image


Looking forward to hearing from you
I hope it matches your vision.
It's a feature that's essential for me and you said it, for other people too :)
If it's ok, I could look at making other plugins compatible.

dh47:
What's the benifit to receiving tone only?

all3kcis:
Personally, I use it to turn on my bedroom lights when my Pager rings at night ;)

dh47:
Ok.

PinPrien:
I also approuve it, as my pager first tone only and I get message 10/20 sec after. Usefull time saved! :)

all3kcis:
I'm ready to merge
@DanrwAU @davidmckenzie

PR #286, opened 2019-05-14
DoS Vulnerability Fix

DanrwAU:
# Description

Another DoS Vulnerability as a result of #278 

Fixes # No open issue

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Confirmed Changes no longer loop

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated changelog.md with changes made

PR #288, opened 2019-05-14
Search looping fix yet again

davidmckenzie:
# Description

Turns out I was _way_ overthinking this. We can get rid of the comparison on states completely if we only push the state when we actually have to. Wrapping it in an `if (page)` completely removes the need to compare states, thus removing the fuckery that is URL encoding.

The state is still encoded to prevent issues with `?` and `&` etc., but there should now be a near 0 chance of looping happening.

Fixes #286

## Type of change

# How Has This Been Tested?

Tried to break it

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 **MySQL/MariaDB Upgrades from 0.3.0/0.3.1 aren't possible. Databases will need to be recreated and data re-imported.**
 
-**Fixes DoS Vulerability in search handling** #286 @davidmckenzie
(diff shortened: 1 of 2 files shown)


davidmckenzie:
Oh and sqlite still errs on certain special chars, but that's a feature not a bug or something.

Just don't search for things with `"` in it

DanrwAU:
> Just don't search for things with `"` in it

Your not the boss of me. 😝 

Good work, i'll do some testing this afternoon and see if i can break it 😁

PR #298, opened 2019-07-26
Admin Show - Disable PDW mode for admins

marshyonline:
# Description

This creates a setting called "Disable PDW Mode for Admins"
It allows logged in users to see the messages stored in the database without alias's but continues to run in PDW mode for the guests

This allows admins to find capcodes they might not have stored quickly and easyly 

## Type of change

# How Has This Been Tested?

Tested by enable/disabling the following: API Sec, Hide Capcodes, PDW Mode
I ran through each mode with different combos to ensure messages still show correctly

# Checklist:

### DIFF
--- a/server/routes/api.js
+++ b/server/routes/api.js
@@ -661,28 +661,40 @@ router.post('/messages', isLoggedIn, function(req, res, next) {
                                             delete row.pluginconf;
                                             if (HideCapcode || apiSecurity) {
                                               //Emit full details to the admin socket
-                                              req.io.of('adminio').emit('messagePost', row);
+                                              if (pdwMode && adminShow) {
+                                                req.io.of('adminio').emit('messagePost', row);
+                                              } else if (!pdwMode) {
+                                                req.io.of('adminio').emit('messagePost', row);
+                                              }
                                               //Only emit to normal socket if HideCapcode is on and ApiSecurity is off.
                                               if (HideCapcode && !apiSecurity) {
-                                                // Emit No capcode to normal socket
-                                                row = {
-                                                  "id": row.id,
-                                                  "message": row.message,
-                                                  "source": row.source,
-                                                  "timestamp": row.timestamp,
-                                                  "alias_id": row.alias_id,
-                                                  "alias": row.alias,
-                                                  "agency": row.agency,
-                                                  "icon": row.icon,
-                                                  "color": row.color,
-                                                  "ignore": row.ignore,
-                                                  "aliasMatch": row.aliasMatch
-                                                };
-                                                req.io.emit('messagePost', row);
+                                                if (pdwMode && !row.aliasMatch) {
+                                                  //do nothing
+                                                } else {
+                                                  // Emit No capcode to normal socket
+                                                  row = {
+                                                    "id": row.id,
+                                                    "message": row.message,
+                                                    "source": row.source,
+                                                    "timestamp": row.timestamp,
+                                                    "alias_id": row.alias_id,
+                                                    "alias": row.alias,
+                                                    "agency": row.agency,
+                                                    "icon": row.icon,
+                                                    "color": row.color,
+                                                    "ignore": row.ignore,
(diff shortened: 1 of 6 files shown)


marshyonline:
@DanrwAU + @davidmckenzie 
I'm unsure of how we can wrangle the POST /messages API to allow the messages to emit via the websocket to admins only
any ideas?

DanrwAU:
> @DanrwAU + @davidmckenzie
> I'm unsure of how we can wrangle the POST /messages API to allow the messages to emit via the websocket to admins only
> any ideas?

I'd suggest a rewrite of api.js to from line 607 onwards to process all messages. 

You'd then be able to emit to the admin socket always and/or filter the normal socket emitting via an `if (pdwMode)`

server/routes/api.js
```diff
@@ -67,13 +68,21 @@ router.get('/messages', isLoggedIn, function(req, res, next) {
     initData.limit = parseInt(defaultLimit, 10);
   }
   if (pdwMode) {
-    var subquery = db.from('capcodes').where('ignore', '=', 0).select('id')
+    if (adminShow && req.isAuthenticated()) {
```
DanrwAU: Probably easier to move this check up to the pdwMode check 

Do 
`if (pdwMode && !req.isAuthenticated())` on line 70 

That way we aren't doubling up the same query as line 72 👍

server/routes/api.js
```diff
@@ -67,13 +68,21 @@ router.get('/messages', isLoggedIn, function(req, res, next) {
...
+    } else {
+      var subquery = db.from('capcodes').where('ignore', '=', 0).select('id')
+    }    
   } else {
     var subquery = db.from('capcodes').where('ignore', '=', 1).select('id')
   }
   db.from('messages').where(function () {
     if (pdwMode) {
-      this.from('messages').where('alias_id', 'in', subquery)
+      if (adminShow && req.isAuthenticated()) {
```
DanrwAU: Same thing here, if you move the check up 1 line you don't have to double up the query 😁

server/routes/api.js
```diff
@@ -106,7 +115,11 @@ router.get('/messages', isLoggedIn, function(req, res, next) {
         })
         .modify(function(queryBuilder) {
           if (pdwMode) {
-            queryBuilder.innerJoin('capcodes', 'capcodes.id', '=', 'messages.alias_id').where('capcodes.ignore', 0)
+            if (adminShow && req.isAuthenticated()) {
```
DanrwAU: And again 😁

davidmckenzie:
Oooh neat!

Yeah for the socket I think you'd have to rip out this bit:

```
                                          if (pdwMode) {
                                            queryBuilder.innerJoin('capcodes', 'capcodes.id', '=', 'messages.alias_id')
                                          } else {
                                            queryBuilder.leftJoin('capcodes', 'capcodes.id', '=', 'messages.alias_id')
                                          }
```

replacing with just `queryBuilder.leftJoin('capcodes', 'capcodes.id', '=', 'messages.alias_id');`

then do the pdw mode logic stuff around the `if (HideCapcode || apiSecurity) {` bit

You might also have to modify this bit: https://github.com/pagermon/pagermon/blob/master/server/views/index.ejs#L259-L283

Probably just `if (hidecapcode || whatevertheadminshowvariablethingis)`

marshyonline:
Have changed the SQL parts but am stuck on the message logic, need to really sit and think about it

DanrwAU:
@davidmckenzie Is that what you had in mind?

We didn't need to modify the index.ejs as the SQL Statement for getting messages was modified in api.js 😁

DanrwAU:
@davidmckenzie we think this one is done, do you wanna review and test for us? 

This might also be a "squash and merge" situation, since i have no professionalism at all :-P

marshyonline:
I've run thu the tests myself but can't approve @DanrwAU 's changes

davidmckenzie:
Tests good here, awesome work \o/

Issue #256, opened 2019-04-12
SQLite - Duplicate Message during alias refresh


DanrwAU:
**Describe the bug**
Duplicate Message checking does not seem to work if messages are submitted while AliasRefresh is running. 

**To Reproduce**
Start long running aliasRefresh cycle and submit duplicate messages

**Expected behavior**
Duplicate messages filtered correctly. 

**Additional context**
Tested only in v0.2.3

davidmckenzie:
I'm surprised messages even submit at all during the refresh tbh

DanrwAU:
Me too, they seem to hold and be processed later ?

DanrwAU:
I guess this could be related to #264 in that they are all submitted at the same time after the aliasRefresh finishes?

DanrwAU:
I haven't seen this one since. closing as fixed i guess?

PR #301, opened 2019-08-09
Optimize Duplicate checking queries

### LINKED ISSUE #264: v0.3.0 - Duplicate Checking not working
**Describe the bug**
Duplicate checking not working when message limit is set to 0

**To Reproduce**
Set duplicate limit to 0, duplicate time to 60. 
Submit multiple copies of the same message (more than 2) 
Only the first is filtered, subsequent messages are submitted

**Expected behavior**
All messages are filtered correctly

**Additional context**
This **MAY** have already been fixed with the commits i've made but intermittently still seeing the issue. 
Issue for visibility. 
This one should block release of 0.3.0 for now.

DanrwAU:
# Description

Possible fix for the duplicate message checking issues introduce in v0.3.0 and Knex when using multiple receivers. 

Doesn't Fix #264 

## Type of change

# How Has This Been Tested?

Running this in one of my production environments and seeing significantly better dupe checking. Would like feedback from others who were seeing this issue before calling it fixed though.

@s3m1s0n1c can you try this one out and see if your still seeing dupes?

# Checklist:

### DIFF
--- a/server/routes/api.js
+++ b/server/routes/api.js
@@ -582,7 +582,7 @@ router.post('/messages', isLoggedIn, function(req, res, next) {
                       .as('temp_tab')
                     })  
               })
-              .andWhere('message', 'like', `%${message}%`)
+              .andWhere('message', '=', message)
               .andWhere('address', '=', address)
             } else if ((dupeLimit !=0) && (dupeTime == 0)) {
               queryBuilder.where('id', 'in', function () {
@@ -595,7 +595,7 @@ router.post('/messages', isLoggedIn, function(req, res, next) {
                           .as('temp_tab')
                     })
               })
-              .andWhere('message', 'like', `%${message}%`)
+              .andWhere('message', '=', message)
               .andWhere('address', '=', address)
             } else if ((dupeLimit == 0) && (dupeTime != 0)) {
               queryBuilder.where('id', 'in', function () {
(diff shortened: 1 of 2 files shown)


davidmckenzie:
Interesting! I'm guessing that handles the fake-escaped characters a little better, unless there's some minor difference in whitespace between the two clients or something that's causing the dupes?

DanrwAU:
Don't know why it worked, but postgres threw a hissy at the way i was doing it.... doing it this way seems to solve both issues *shrug*

DanrwAU:
Dammit 
image

s3m1s0n1c:
@DanrwAU Copy, Will try and give it a go as soon as I can..

Thanks

DanrwAU:
Possibly got a little bit excited, i'm still seeing *some* nowhere near as many as before tho 
image

server/routes/api.js
```diff
@@ -595,18 +595,18 @@ router.post('/messages', isLoggedIn, function(req, res, next) {
                           .as('temp_tab')
                     })
               })
-              .andWhere('message', 'like', message)
+              .andWhere('message', 'like', `%${message}%`)
```
DanrwAU: Thinking about this further, do these even need to be like's? the whole idea is we want to match exact messages anyway. Thoughts?

davidmckenzie: Yknow... I think you might be onto something. If it's = instead of LIKE then it might do less dodgy shit with special chars in the string. And from my Googling it appears there's more weirdness with LIKE too: https://stackoverflow.com/questions/543580/equals-vs-like

DanrwAU: Pushed that change to my dupe heavy instance, looking good so far, see how we go

DanrwAU: I'd be curious if it makes a difference for #256 as well

DanrwAU:
I am definetly still getting some duplicates so not a great fix. I think these changes should stay in anyway but they don't have the desired effect 😭

DanrwAU:
I think the dupes come down to the changes made with Knex, the previous driver was serializing and this one isn't. 

@marshyonline and I were discussing implementing a temp table and some sort of queueing as a possible solution but that sounds hard.... 😭

PR #305, opened 2019-08-14
WIP Knex Postgres Support

DanrwAU:
# Description

Add support for Postgres through Knex

## Type of change

# How Has This Been Tested?

lol nope

# Checklist:

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -38,6 +38,7 @@
     "passport": "^0.3.2",
     "passport-local": "^1.0.0",
     "passport-localapikey-update": "^0.5.0",
+    "pg": "^7.12.0",
     "pmx": "latest",
     "pushover-notifications": "latest",
     "request": "latest",
--- a/server/public/templates/admin/settings.html
+++ b/server/public/templates/admin/settings.html
@@ -43,8 +43,8 @@
                 <select name="database.type" ng-model="settings.database.type" class="form-control" required>
                   <option value="sqlite3">SQLite</option>
                   <option value="mysql">MySQL / MariaDB</option>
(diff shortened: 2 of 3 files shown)


davidmckenzie:
> I cant remember why i did this possibly MySQL?

It's from the dodgy wildcard handling in aliases, since the aliases are strings containing wildcards, it means the order of typical matching has to be reversed - you have to say `if 101000 is like 101___` instead of `if 101___ = 101000` or something like that. "string" must always be on the left and "pattern" must always be on the right.

The particular way we're doing it might be some sqlite shorthand, not sure... if all else fails, postgres has much more powerful regex that we could use to our advantage - https://www.postgresql.org/docs/9.3/functions-matching.html

DanrwAU:
> > I cant remember why i did this possibly MySQL?
> 
> It's from the dodgy wildcard handling in aliases, since the aliases are strings containing wildcards, it means the order of typical matching has to be reversed - you have to say `if 101000 is like 101___` instead of `if 101___ = 101000` or something like that. "string" must always be on the left and "pattern" must always be on the right.
> 
> The particular way we're doing it might be some sqlite shorthand, not sure... if all else fails, postgres has much more powerful regex that we could use to our advantage - https://www.postgresql.org/docs/9.3/functions-matching.html

That does ring a bell! 

That explains that then this is going to be painful making them all play nicely together...
I was hoping postgres wouldn't be too much of a stretch from the other 2 but it's looking quite a bit more difficult 😭

DanrwAU:
Closing this, these changes won't work, will need a rethink

PR #325, opened 2019-09-26
Add Webhook plugins

stubbers:
# Description

Add 2 plugins

## Simple Webhook

User has checkboxes for each of the fields (alias, agency, capcode, message, timestamp, etc)
Content-type locked to JSON
Payload generated as a JSON object by the plugin
Plugin can be set to send all messages, all defined aliases, or the usual specific aliases

## Advanced Webhook

User specifies payload 'template' (ie - "{message: '/message/'}") and the plugin replaces /message/ with the message (alias, agency, capcode, message, timestamp, etc)
User specifies content-type in a text field
Plugin can be set to send all messages, all defined aliases, or the usual specific aliases

## Type of change

# How Has This Been Tested?

Tested using a node-red instance using HTTP and a custom port
Both Advanced and Simple plugins were tested in filtered mode and non-filtered mode

# Checklist:


stubbers:
Feedback very welcome, haven't used git in a while...

davidmckenzie:
Awesome! Some thoughts - the payload it sends seems a bit limited, would it be better to send all of the raw data as individual strings in the object (alias, agency, capcode, message, timestamp, etc) and let the receiving end construct it in the way it wants? Would keep it more generic that way

eopo:
I’d rather define the format of the payload by plugin config, so one can use any API, not just custom ones

marshyonline:
I have to agree with @eopo 
Make what you are sending to the webhook customisable via config/settings

stubbers:
Thanks for the feedback, I agree I'd prefer it was extensible. I was just going for a quick fix but that could cause breaks down the line if I (or someone) made it extensible.

Leave it with me and I'll see what I can do!

stubbers:
I propose splitting this into 2 different plugins:

# Simple Webhook

User has checkboxes for each of the fields (alias, agency, capcode, message, timestamp, etc)
Content-type locked to JSON
Payload generated as a JSON object by the plugin

# Advanced Webhook

User specifies payload 'template' (ie - "{message: '/message/'}") and the plugin replaces /message/ with the message (alias, agency, capcode, message, timestamp, etc)
User specifies content-type in a text field

Happy to do both under this pull just want to ensure I'm not spending time on something that'll just get knocked back

eopo:
I’d rather make version 1 as a possible template to version 2

DanrwAU:
> I propose splitting this into 2 different plugins:
> 
> # Simple Webhook
> User has checkboxes for each of the fields (alias, agency, capcode, message, timestamp, etc)
> Content-type locked to JSON
> Payload generated as a JSON object by the plugin
> 
> # Advanced Webhook
> User specifies payload 'template' (ie - "{message: '/message/'}") and the plugin replaces /message/ with the message (alias, agency, capcode, message, timestamp, etc)
> User specifies content-type in a text field
> 
> Happy to do both under this pull just want to ensure I'm not spending time on something that'll just get knocked back

I like this approach 👍

eopo:
Do we have a chance of implementing plugin instances, so we can use multiple webhooks without copying the code

davidmckenzie:
> I like this approach 👍
Agreed, though don't feel obliged to do both if you only need one, we can always put it on the backlog to come back to :)

> Do we have a chance of implementing plugin instances, so we can use multiple webhooks without copying the code

I can see a way of doing this but would require significant refactor, will chuck an issue in for it

stubbers:
@DanrwAU Ready for another look. Have split it into 2 and added a feature to allow all messages or all messages with aliases to be sent without needing a checkbox under each individual alias

server/plugins/AdvancedWebhook.json
```diff
@@ -0,0 +1,50 @@
...
+        {"value": "1", "text": "Defined aliases"},
+        {"value": "2", "text": "All messages"}
+      ],
+      "required": true
+    },
+    {
+      "name": "contentTemplate",
+      "label": "Content Template",
+      "description": "Template of content to send. /<field>/ will be replaced with the contents for the message (eg - /message/ will get replaced with the message text)",
+      "type": "text",
```
davidmckenzie: I should probably add a textarea field type available for this... will add an issue

server/plugins/AdvancedWebhook.json
```diff
@@ -0,0 +1,50 @@
...
+        {"value": "0", "text": "Per alias only"},
+        {"value": "1", "text": "Defined aliases"},
+        {"value": "2", "text": "All messages"}
+      ],
+      "required": true
+    },
+    {
+      "name": "contentTemplate",
+      "label": "Content Template",
+      "description": "Template of content to send. /<field>/ will be replaced with the contents for the message (eg - /message/ will get replaced with the message text)",
```
davidmckenzie: May want to provide a list of available fields so users don't have to reference code if they forget syntax

eopo: Gonna throw in Model layer there. This way we could pull available fields from there, including plugin-generated ones 😏

stubbers: I can add it to the description but was wary of the length. I can write a documentation page but there isn't really a place to put one yet :-P

Let me know if you want it added to the description

DanrwAU:
@pagermon/collaborators what's your thoughts on this one, ready to go or more changes required?

davidmckenzie:
I'm a terrible person and forgot about this, happy for it to merge as is - @stubbers wanna add docos to wiki somewhere as well?

PR #358, opened 2020-03-21
fix regex maybe

### LINKED ISSUE #357: Flex no Longer Being Sent to Pagermon Server
It seems multimon-ng has changed the way Flex is phased.

Inkedunknown_LI

This is what Flex looks like now just via multimon-ng.

`FLEX|2020-02-26 06:46:15|1600/2/K/A|11.081|001234567|ALN|This is a test periodic page. 95 19:46`

A few people on the discord has advised that the format expected is different in the reader.js file.

Anyone else can chuck in there 2cents, please?  Ta.

davidmckenzie:
# Description

Should make the regex work with both old and new formats. My testing was limited to throwing a few test cases through it so I have no idea how it will perform in the real world. There's a high probability this will break things for FLEX users

Fixes #357

## Type of change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,3 +93,4 @@ server/sessions
 
 client/samples/testymctest.sh
 
+.vscode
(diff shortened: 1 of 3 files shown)


dh47:
Just tested this and so far it checks out.

PR #384, opened 2020-04-25
Add ability to select themes, support additional themes

DanrwAU:
# Description

Adds the ability to add and select themes without the need to modify core files. 

image

Themes can be added by copying the "Default" folder to a new folder inside the themes directory, then modifying as required.

## Type of change

# How Has This Been Tested?

ITS WORKING

# Checklist:


DanrwAU:
@pagermon/collaborators if you could look this one over. 

I don't think i've introduced any glaring security holes but never know 🤣

server/routes/admin.js
```diff
@@ -75,8 +75,12 @@ router.route('/settingsData')
                     plugins.push(pConf);
             }
         });
+        let themes = [];
+        fs.readdirSync('./themes').forEach(file => {
```
SloCompTech, line 79: Don't hardcode path, use path.join with __dirname or something so it is relative to this file, not CWD

DanrwAU: Done that way as that's how the Plugins are done!
I blame @davidmckenzie

SloCompTech: Please just fix it, because this can result in some bugs if you are running server from different cwd than expected, it is best to make this robust as possible

DanrwAU: Hmmm replacing this line with `fs.readdirSync(path.join(__dirname, 'themes'))` breaks settings loading completely... 
Any ideas?

DanrwAU: Hmmm no i'm not... 

replacing this line with `fs.readdirSync(path.join(__dirname, 'themes')).forEach(file => {` doesn't work. 😢

SloCompTech: You sure, because here `path.join(__dirname, 'themes')` gives expected result. Where is `themes` directory ?

DanrwAU: Yup, i don't think it likes being called inside fs.readdirSync 

It works fine if i throw it out to the logger :|

PR #392, opened 2020-05-03
Add CSV Import to capcodes 

### LINKED ISSUE #327: CSV import for Aliasses
Hey, I'd recommend implementing CSV import for aliasses, the way with import.js has lots of possible errors and need a lot of manual work.
Implementing a CSV import where the end user can select which field goes to which and if existing aliasses for the same address should be ignored or overridden would be best.

DanrwAU:
# Description

Adds the ability to import CSV files for capcodes. 

Requires a header row containing at least address & alias all other fields are optional. 

There is currently no filtering for invalid headers and these are ignored (YOU HAVE BEEN WARNED!)

@pagermon/collaborators this one will need an actual review, I'm sure my code is dodgy to achieve what I have. 

Fixes #327 
## Type of change

- [x] This change requires a documentation update

# How Has This Been Tested?

Importing CSV's with correct header row works as expected
Trying to import CSV with no header row fails. 

# Checklist:

### DIFF
--- a/server/routes/api.js
+++ b/server/routes/api.js
@@ -1007,26 +1007,26 @@ router.post('/capcodeRefresh', isLoggedIn, function(req, res, next) {
   })
 });
 
-router.post('/capcodeExport', isLoggedIn, function(req, res, next) {
+router.post('/capcodeExport', isLoggedIn, function (req, res, next) {
   nconf.load();
   var dbtype = nconf.get('database:type');
   var filename = 'export.csv'
   db.from('capcodes')
     .select('*')
-    .modify(function(queryBuilder) {
+    .modify(function (queryBuilder) {
       if (dbtype == 'oracledb')
         queryBuilder.orderByRaw(`REPLACE("address", '_', '%')`);
       else
         queryBuilder.orderByRaw(`REPLACE(address, '_', '%')`)
     })
     .then((rows) => {
-      converter.json2csv(rows, function(err, data) {
+      converter.json2csv(rows, function (err, data) {
         if (err) {
           res.status(500).send(err);
         } else {
           res.status(200).send({ 'status': 'ok', 'data': data })
         }
-      }) 
+      })
     })
     .catch((err) => {
       logger.main.error(err);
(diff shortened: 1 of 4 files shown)


server/routes/api.js
```diff
@@ -1007,33 +1007,137 @@ router.post('/capcodeRefresh', isLoggedIn, function(req, res, next) {
...
+          var color = capcode.color || 'black';
+          var icon = capcode.icon || 'question';
+          var ignore = capcode.ignore || 0;
+          var pluginconf = JSON.stringify(capcode.pluginconf) || "{}";
+          await db('capcodes')
+            .returning('id')
+            .where('address', '=', address)
+            .first()
+            .then((rows) => {
+              if (rows) {
```
SloCompTech, line 1063: Add comment: Update alias with existing address

server/routes/api.js
```diff
@@ -1007,33 +1007,137 @@ router.post('/capcodeRefresh', isLoggedIn, function(req, res, next) {
...
+                    })
+                  })
+                  .catch((err) => {
+                    importresults.push({
+                      address: address,
+                      alias: alias,
+                      result: 'failed' + err
+                    })
+                  })
+              } else {
```
SloCompTech, line 1089: Add comment: Create new alias (address not in db)

server/routes/api.js
```diff
@@ -1007,33 +1007,137 @@ router.post('/capcodeRefresh', isLoggedIn, function(req, res, next) {
...
     .catch((err) => {
       logger.main.error(err);
       return next(err);
     })
 });
 
+router.post('/capcodeImport', isLoggedIn, function (req, res, next) {
+  for (var key in req.body) {
+    //remove newline chars from dataset
+    req.body[key] = req.body[key].replace(/\r/g, '');
```
SloCompTech, line 1040: Could here be: `req.body[key] = req.body[key].replace(/[\r\n]/g, '');` ?

DanrwAU: > req.body[key] = req.body[key].replace(/[\r\n]/g, '');

Does this replace both \r and \n individually?

SloCompTech: yes, as far as I remember it does (`[]` is for specifying a group of chars that match in regex), but better to test it to be sure

PR #403, opened 2020-05-07
Fix Multi-Delete Capcodes on MYSQL

DanrwAU:
# Description

In some circumstances on MySQL, using the multidelete can throw an error

```
"Cannot delete or update a parent row: a foreign key constraint fails (`pagermon-test`.`messages`, CONSTRAINT `messages_alias_id_foreign` FOREIGN KEY (`alias_id`) REFERENCES `capcodes` (`id`))"
```

This recreates with column with an SET NULL on delete key.

This requires dropping the alias_id column first, which can take quite some time on a large database.
It will also require an alias refresh once the process is complete. 

Fixes # No issue created

## Type of change

# How Has This Been Tested?

Tested Migrations on copy of my production database.

# Checklist:

### DIFF
--- a/server/knex/migrations/20200507165756_messages_foreignkey_fix.js
+++ b/server/knex/migrations/20200507165756_messages_foreignkey_fix.js
@@ -11,10 +11,14 @@ exports.up = function (db, Promise) {
                     table.dropColumn('alias_id')
                 })
                     .then(function () {
-                        return db.schema.table('messages', function (table) {
+                        return db.schema.table('', function (table) {
                             table.integer('alias_id').unsigned().references('id').inTable('capcodes').onDelete('SET NULL');
+                            
+                        })
+                        .then(function () {
                             nconf.set('database:aliasRefreshRequired', 1);
                             nconf.save();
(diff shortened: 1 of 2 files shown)


server/knex/migrations/20200507165756_messages_foreignkey_fix.js
```diff
@@ -11,10 +11,14 @@ exports.up = function (db, Promise) {
                     table.dropColumn('alias_id')
                 })
                     .then(function () {
-                        return db.schema.table('messages', function (table) {
+                        return db.schema.table('', function (table) {
```
SloCompTech, line 14: Why is table name missing ?

DanrwAU:
@pagermon/collaborators i'm not aware of a better way of doing this than dropping the column and recreating, it doesn't seem to be alterable.

Let me know if you know a better way 👍

DanrwAU:
I ran this against my database with 2.5mil messages and 2000 aliases, it was not a fun time ... i can't imagine it would be any fun running that on sqlite either tho 😁

PR #407, opened 2020-05-10
Fix Shell plugin - Missing Callback

all3kcis:
The message was not displayed automatically (websocket broken).

# Description

Fix Shell plugin - Missing Callback

## Type of change

### DIFF
--- a/server/plugins/Shell.js
+++ b/server/plugins/Shell.js
@@ -49,6 +49,8 @@ function run(trigger, scope, data, config, callback) {
     }else{
       logger.main.info('File '+full_file_name+' not exist');
     }
+
+    callback(data);
   }
 }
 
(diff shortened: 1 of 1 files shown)


server/plugins/Shell.js
```diff
@@ -49,6 +49,8 @@ function run(trigger, scope, data, config, callback) {
     }else{
       logger.main.info('File '+full_file_name+' not exist');
     }
+
+    callback(data);
   }
```
SloCompTech, line 54: Missing
```
} else {
  callback();
}
```
for the main if in function, because plugin must always call `callback`
See [here](https://github.com/pagermon/pagermon/tree/v0.3.9/server/plugins).

SloCompTech:
And also add record to changelog

DanrwAU:
If your happy with this without the changelog @SloCompTech it looks like this was based off a slightly older fork, i'll fix the changelog upstream

PR #413, opened 2020-05-17
WIP adds geolocation features

ryanunderwood:
# Description

WIP - To automatically filter and normalize address data (grid ref, full addresses, co-ordinates) into standard WGS84 co-ordinates from incoming pager messages. User-defined Regex and expected address type from that regex output (full_address, grid_references etc.). 
At time of pull request config.json not included but will be in further commits. Also implements geoforwarding API (at time of PR this is LocationIQ, a bit hacky at this time. Hope to include expected Country, City etc. to select the correct address output).

Works on an Agency basis to cover multiple capcodes for the same service (ambulance, fire etc.).  Requires a column addition in the DB, which will be included in the branch soon.

Automatic mapping based on normalized co-ordinates from addresses, filterable. Works on an "after" basis as to have access to agency and not break anything.

Lots of debug stuff in there atm oops

# How Has This Been Tested?

It's working well for me. Still needs work.

# Checklist:


client/config/default.json
```diff
@@ -1,7 +0,0 @@
-{
```
DanrwAU, line 1: Why was this file removed? This can't be merged

client/reader.sh
```diff
@@ -1,25 +0,0 @@
-#!/bin/bash
```
DanrwAU, line 1: Again, this shouldn't be in your PR

server/app.js
```diff
@@ -1,4 +1,4 @@
-var version = "0.3.8-beta";
```
DanrwAU, line 1: Again - this is only touched by version commits, shouldn't be here

server/routes/api.js
```diff
@@ -574,11 +574,15 @@ router.post('/messages', isLoggedIn, function(req, res, next) {
...
         return res.send('Ignoring filtered');
       }
+
+      // logger.main.error('Say hello to API.JS - IS RAW GEOLOCATION HERE?! Following: ' + JSON.stringify(data));
         var address = data.address || '0000000';
         var message = data.message || 'null';
         var datetime = data.datetime || 1;
         var timeDiff = datetime - dupeTime;
         var source = data.source || 'UNK';
+        var raw_geolocation = data.raw_geolocation || 0;
```
DanrwAU, line 584: The whole point of the plugin system is not to clutter up API.js from extra features like this not everybody is going to use. 
Plugins shouldn't touch API.js in my opinion

server/routes/index.js
```diff
@@ -60,10 +63,41 @@ router.get('/testLogin', isLoggedIn, function(req, res) {
     });
     
 /* GET home page. */
-router.get('/', function(req, res, next) {
+router.get('/', function(req, res, next) {7
   res.render('index', { pageTitle: 'Home' });
 });
 
+router.get('/map', function(req,res,next) {
```
DanrwAU, line 70: Not really a fan of this either, sort of jumping out of the realms of a plugin and into the framework of the underlying app. 
In saying that i don't know how to make it better either, @pagermon/collaborators any ideas?

eopo:
I’d say best way would be to implement handling existing geodata into the core app (Display, filtering, Maps link Generation, ...) and leave filling these fields to custom plugins to add the needed flexibility regarding different formats of messages and so on ...

ryanunderwood:
Perhaps to extend the plugin capabilities, the api.js may run a separate loader to ensure certain enabled variables needed for the plugins are present. As well as allowing plugin routes, for instance where the /map route sits in the underlying app, maybe loading these in from the plugin "package" as it were, to allow a clean separation?

server/routes/index.js
ryanunderwood: Ultimately was a quick put-together for the sake of making it work. Would appreciate input on how to improve the overall structure, and better ways to implement more advanced plugins in general.

client/config/default.json
ryanunderwood: My bad. Accidentally committed changes that shouldn't have been in there. Didn't mean to delete them from the branch entirely :/

davidmckenzie:
> Perhaps to extend the plugin capabilities, the api.js may run a separate loader to ensure certain enabled variables needed for the plugins are present. As well as allowing plugin routes, for instance where the /map route sits in the underlying app, maybe loading these in from the plugin "package" as it were, to allow a clean separation?

That's really pushing beyond what the plugin system is designed for, and opens up a lot more questions about security etc.

Could be a really cool feature to merge in with main as long as it's flexible enough to be used by everyone :) Having a map tab that shows incidents in the last `n` time would be super neat.

server/plugins/RawGeoFromRegex.js
```diff
@@ -0,0 +1,121 @@
...
+        })
+        .catch ((err) => {
+            logger.main.error('GEO DB ERROR = ' + err);
+        });
+    // logger.main.error('filtered message = ' + data.raw_geolocation);
+    return coords;
+}
+
+function getCoordsFromAddress(address, apiKey, messageID) {
+    getJSON('https://eu1.locationiq.com/v1/search.php?key=' + apiKey + '&q=' + address +'&format=json',
```
davidmckenzie, line 75: The getJSON lib seems unnecessary here - there was also a commit ahead of your branch that moved all plugins to use `axios`, so probably best to switch to using that

server/routes/index.js
davidmckenzie: Ideally the route would be in the frontend js, and would pull data from API endpoint

DanrwAU:
Closing as abandonned. Original Repo no longer exists, none of the reuqested changes have been made

PR #458, opened 2021-01-29
Improved wildcard feature

eopo:
If a match between message and alias was due to a wildcard,
a new row "wildcard" is returned. In the message popover,
the user is now asked if he wants to create a new alias using the message,
if the match was using a wildcard.

# How Has This Been Tested?

- [X] Own testing system
- [X] Own production system
- [ ] Request testing on another system!!!

# Checklist:

### DIFF
--- a/server/routes/api.js
+++ b/server/routes/api.js
@@ -112,7 +112,7 @@ router.route('/messages')
           var rowCount
 
           db.from('messages')
-            .select('messages.*', 'capcodes.alias', 'capcodes.agency', 'capcodes.icon', 'capcodes.color', 'capcodes.ignore')
+            .select('messages.*', 'capcodes.alias', 'capcodes.agency', 'capcodes.icon', 'capcodes.color', 'capcodes.ignore', 'CASE WHEN NOT capcodes.address = messages.address THEN 1 ELSE 0 END as wildcard')
             .modify(function (queryBuilder) {
               if (pdwMode) {
                 if (adminShow && req.isAuthenticated() && req.user.role == 'admin') {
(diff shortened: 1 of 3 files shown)


davidmckenzie:
Oh that's neat logic, I like it

DanrwAU:
Oooo I've tried to fix this a few times and could never solve this... Nice work

davidmckenzie:
No idea whether that will work with other DB types... if it does fail you could just include capcodes.address and include the logic in UI (`if capcodes.address index of % or * or whatever`)

eopo:
Fixed two bugs I had, does work on both of my systems (develop state with this change and master state with heavy customizations) now. I'd like to have one of you try it as well ;)

eopo:
> No idea whether that will work with other DB types... if it does fail you could just include capcodes.address and include the logic in UI (`if capcodes.address index of % or * or whatever`)

I checked the syntax of CASE, should be the same for SQLITE and OracleDB according to their docs. Would be nice if someone running sqlite and someone running oracle could double check that tho.

DanrwAU:
>  someone running oracle could double check that tho.

Wait, there is someone crazy enough to run Orcale? 

I'll upgrade my dev box later and see how it goes on MySQL and SQLite :)

davidmckenzie:
> > someone running oracle could double check that tho.
> 
> Wait, there is someone crazy enough to run Orcale?
> 
> I'll upgrade my dev box later and see how it goes on MySQL and SQLite :)

Just me on my dev box that I never cut over :joy: I likely won't get around to testing this on there so feel free to merge if MySQL and SQLite work

eopo:
Hmm. I'm thinking it might actually be better to store the wildcard match field permanently with the message, to improve the alias rematching process. It could then just re-parse all messages with wildcard fielt set true or that are linked to an alias that was changed.
New table that stores alias id of aliases that have been changed. Every time the alias match is re-run it rematches all messages with wildcard field and all messages, whose alias id is contained in the changed aliases table.

DanrwAU:
This seems to be working ok for me. 

Only issue i have is the UI is kinda ugly 🤣 

image

DanrwAU:
Much betterer
image

DanrwAU (approved):
Test on SQLite and MySQL, seems to work correctly. 

I'm happy to merge this one to be included in 0.3.11

eopo:
🎉

PR #462, opened 2021-02-26
Migration script

RIKIKU:
# Description

Adds a crude script to copy messages from a pagermon Sqlite DB to an instance of Pagermon. 

Fixes # none

## Type of change

# How Has This Been Tested?

I migrated my sqlite db to an instance of Pagermon running mariadb. ~250K messages.

# Checklist:

- [ ] I have updated changelog.md with changes made. 
No change to the application code.

### DIFF
--- a/scripts/MigrateMessages.ps1
+++ b/scripts/MigrateMessages.ps1
@@ -44,14 +44,6 @@ class message {
         $this.Alias_Id = $input.Alias_Id
     }
 }
-$GetHighestId = @"
-SELECT Id
-FROM messages
-ORDER BY Id Desc
-limit 1
-"@
-$DbConnection = New-SQLiteConnection -DataSource $DbPath -ErrorAction Stop
-$maxRows = Invoke-SqliteQuery -SQLiteConnection $DbConnection -Query $GetHighestId -ErrorAction Stop
 
 function New-PagerMonMessage {
     #send a message to pagermon.
@@ -81,6 +73,15 @@ function resumeFrom ($indexPath) {
     Write-Output $output
 }
 #endregion
+
+$GetHighestId = @"
+SELECT Id
+FROM messages
+ORDER BY Id Desc
+limit 1
+"@
+$DbConnection = New-SQLiteConnection -DataSource $DbPath -ErrorAction Stop
+$maxRows = Invoke-SqliteQuery -SQLiteConnection $DbConnection -Query $GetHighestId -ErrorAction Stop
 
 $indexPath = "$PSScriptRoot\index.txt"
 if (! (Test-Path -Path $indexPath)) {
--- a/scripts/MigrateMessages.ps1
+++ b/scripts/MigrateMessages.ps1
@@ -4,6 +4,8 @@
     Migrate messages from a pagermon Sqlite database to another pagermon server instance.
 .DESCRIPTION
     This script reads the pagermon messages from an sqlite database and sends them to a pagermon server instance.
+    Script can be run multiple times from the same db.
+    Journal file is used to keep track of which messages were migrated.
 .NOTES
     This script does not migrate configuration, users, etc.
 #>
--- a/scripts/MigrateMessages.ps1
+++ b/scripts/MigrateMessages.ps1
@@ -48,8 +48,8 @@ FROM messages
 ORDER BY Id Desc
 limit 1
 "@
-$DbConnection = New-SQLiteConnection -DataSource $DbPath
-$maxRows = Invoke-SqliteQuery -SQLiteConnection $DbConnection -Query $GetHighestId
+$DbConnection = New-SQLiteConnection -DataSource $DbPath -ErrorAction Stop
+$maxRows = Invoke-SqliteQuery -SQLiteConnection $DbConnection -Query $GetHighestId -ErrorAction Stop
 
 function New-PagerMonMessage {
     #send a message to pagermon.
@@ -81,6 +81,9 @@ function resumeFrom ($indexPath) {
 #endregion
 
 $indexPath = "$PSScriptRoot\index.txt"
+if (! (Test-Path -Path $indexPath)) {
+    $null | Out-File -FilePath $indexPath
+}
 
 for ($i = resumeFrom $indexPath; $i -lt $maxRows.Id + 1; $i++) {
     Write-Progress -Activity "Inserting Message" -Status "Message $i of $($maxRows.id) $($i / $maxRows.Id * 100)% Complete: " -PercentComplete ($i / $maxRows.Id * 100)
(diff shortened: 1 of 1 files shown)


scripts/MigrateMessages.ps1
```diff
@@ -0,0 +1,114 @@
...
+$maxRows = Invoke-SqliteQuery -SQLiteConnection $DbConnection -Query $GetHighestId -ErrorAction Stop
+
+function New-PagerMonMessage {
+    #send a message to pagermon.
+    param (
+        [message]$Message,
+        [UriBuilder]$Uri,
+        [string]$apikey
+    )
+    #"https://pagermon.kyledoestech.com:443/api/messages" #http://www.example.com:3000/api/messages
```
marshyonline, line 63: Please remove this URL

marshyonline (changes requested):

scripts/MigrateMessages.ps1
RIKIKU: Face palm. Removed Ty.

geelongmicrosoldering:
Hey, this is great! Thanks!

I was going to try to manually merge the two databases i had, this saved me the trouble.

catSIXe:
https://github.com/techouse/sqlite3-to-mysql works fine tbh

marshyonline (dismissed):

marshyonline (approved):

marshyonline:
@davidmckenzie Thoughts?

DanrwAU:
> @davidmckenzie @DanrwAU Thoughts?

It worries me merging stuff like this into master. 

We've been burnt many times in the with people expecting support on side tools like this..

geelongmicrosoldering:
> > @davidmckenzie @DanrwAU Thoughts?
> 
> It worries me merging stuff like this into master.
> 
> We've been burnt many times in the past with people expecting support on side tools like this..

Valid concern.

You didn't ask for my input, but I can see a lot of other ways this could be done that would be easier, like a bash version that pipes sqlite dump back into node after making a backup (backup, dump/pipe, get user to confirm and prompt to revert changes). For sqlite > sqlite you could literally just 

I feel like a better migration script would be sqlite > mysql. I had to work that one out

geelongmicrosoldering:
> > @davidmckenzie @DanrwAU Thoughts?
> 
> It worries me merging stuff like this into master.
> 
> We've been burnt many times in the past with people expecting support on side tools like this..

Valid concern.

I can see a lot of other ways this could be done that would be easier, like a bash version that pipes sqlite dump back into node after making a backup (backup, dump/pipe, get user to confirm and prompt to revert changes). Or better, just inserts the data straight into the table without triggering any of the plugins or involving node at all..

For sqlite > sqlite you could literally just:

`#!/bin/bash
import_db_file="messages_to_import.db"
current_new_db_file="messages.db"
backup_file="messages.backup.db"


if ! command -v sqlite3 &> /dev/null
then
    echo "sqlite3could not be found"
    exit
fi

echo "Backing up $current_new_db_file to $backup_file"
cp $current_new_db_file $backup_file

echo "Importing messages from $import_db_file to $current_new_db_file..."
echo "ATTACH '$import_db_file' AS backupdb; INSERT INTO messages (address, message, source, timestamp) SELECT address, message, source, timestamp FROM backupdb.messages;" | sqlite3 $current_new_db_file

echo "Refreshing Alias in $current_new_db_file..."
echo "update messages set alias_id = (select id from capcodes where messages.address like capcodes.address order by REPLACE(address, '_', '%') DESC limit 1);" | sqlite3 $current_new_db_file

echo "Done. Please check the result. If there are issues, a backup of $current_new_db_file can be found in $backup_file"
`

Then it only takes a moment, no plugins are triggered, little overhead, and alias refresh is handled. ID should increment as normal in the process and there should be no conflicts... I think..

I feel like a better migration script would be sqlite > mysql. More specifically, the automatic creation of a mysql database with the correct structure and optional import of an sqlite messages file.  I had to work that one out and it was not fun.

davidmckenzie:
Probably best way to go would be to link to this as a Gist in documentation somewhere rather than crowd the repo. I have literally 0 powershell skills so can't validate the scripts safety or anything like that.

Maybe a page in the wiki that links to 3rd party scripts and snippets like this.

RIKIKU:
I actually agree @DanrwAU on this.  Adding it will give an air of official-nes to it and migration scripts tend to need upkeep. Since there's no automated testing for it the script will eventually become inadequate, like if a new column is added to the message table the script will just ignore it. 

The user really needs to have a solid understanding of what it's doing before they run it, and the expectation is they would turn off any plugins that might get in the way before running it. 

A gist and a link to the documentation might be better, but at the same time, out of date, non-functioning documentation is probably worse than no documentation. 

The reason I wrote it was because I wanted to move to mariaDb from sqlite and I couldn't find any info on how to go about doing that with off the shelf tools. I note that others have

RIKIKU:
I actually agree @DanrwAU on this. Adding it will give an air of official-nes to it and migration scripts tend to need upkeep. Since there's no automated testing for it the script will eventually become inadequate, like if a new column is added to the message table the script will just ignore it.

The user really needs to have a solid understanding of what it's doing before they run it, and the expectation is they would turn off any plugins that might get in the way before running it.

A gist and a link to the documentation might be better, but at the same time, out of date, non-functioning documentation is probably worse than no documentation.

geelongmicrosoldering:
I've been thinking about this (overthinking?), and was wondering..

What if the requirement for a 3rd party script to be included was for it to perform some checks before proceeding? Lets say there was a scripts folder, or a "third_party" folder, or individual gist's (it all works) that contained a readme that explained that scripts are community provided and would not be supported at all.

Then the script checks (for example):
- the pagermon version is one of the versions it has been tested on
- any dependencies are satisfied (in the example above there is an example, where i confirm sqlite3 in installed)
Otherwise the script will not proceed.

Also basic usage included at the top of the script (@RIKIKU has this from memory, I tested his script).
This would all be fairly easy to implement into bash, poweshell, php, js, ruby, python, or potentially any other language used.

Provided that is that the version can be obtained from a public url without auth.. Could always add a router.route('/info'). I know that currently I can grab the auth page and filter out the version from there provided the footer wasn't changed but it's hardly ideal.

**My thought process is:** If a script written for version X and Y is executed on version Z, it simply does nothing but inform the user "hey, this only works for version X-Y". That could mitigate some "this script didn't work" issues being raised, because the script is telling the user exactly why it didn't work.
Also, IF a change is made that effects the function of a script (like an extra table column), the script could be updated to support the new version, and could execute slightly different operations depending on the version accordingly.

Just some rambling, just a thought :)

DanrwAU:
The plan over the past few years has been to split the server and clients into seperate repositories. 

Multiple things have made this difficult so far, however i'd be inclined to create a new Repo in the org for user scripts for putting this sort of thing. 
That was we can tag them with versions etc, but keeps them from being pulled down with every copy.

PR #473, opened 2021-06-01
Make mysql cron adjustable from settings

geelongmicrosoldering:
# Description

After switching to mysql(mariadb) I noticed frequently while adding alias, that the database was locked. It looks like this is caused by a cron job refreshing alias every 30 minutes if a refresh is required (I'm playing with the aliases, so it is). My database is reasonably sized (200k messages?) and runs on limited hardware (Raspberry PI 3).

I considered just adjusting the cron to not refresh as often, but I decided to just add the cron job interval settings into the Settings Template instead.

Note: Probably a better way to do this. Also considered using prettyCron to display human readable translation of cron, but decided to just link cron doc instead and label it advanced

cron_settings

Fixes # cron job interval cannot be adjusted easily by admins

## Type of change

- [x] This change requires a documentation update

# How Has This Been Tested?

- [x] Added to my current server instance

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -202,7 +202,7 @@ if (dbtype == 'mysql') {
   // Get CRON from config
   var cronartime = nconf.get('database:aliasRefreshInterval');
   //If value is falsy (undefined, empty, null etc), set as default
-  if (!cronartime){cronartime = "0 5,35 * * * *"}
+  if (!cronartime){cronartime = "0 5,35 * * * *";}
   var aliasRefreshJob = require('cron').CronJob;
   new aliasRefreshJob(cronartime, function() {
     var refreshRequired = nconf.get('database:aliasRefreshRequired')
--- a/server/app.js
+++ b/server/app.js
@@ -199,7 +199,10 @@ app.use(function(err, req, res, next) {
 // Add cronjob to automatically refresh aliases
 var dbtype = nconf.get('database:type')
 if (dbtype == 'mysql') {
+  // Get CRON from config
   var cronartime = nconf.get('database:aliasRefreshInterval');
+  //If value is falsy (undefined, empty, null etc), set as default
+  if (!cronartime){cronartime = "0 5,35 * * * *"}
   var aliasRefreshJob = require('cron').CronJob;
   new aliasRefreshJob(cronartime, function() {
     var refreshRequired = nconf.get('database:aliasRefreshRequired')
--- a/server/app.js
+++ b/server/app.js
@@ -206,7 +206,7 @@ if (dbtype == 'mysql') {
   //If value is falsy (undefined, empty, null etc), set as default
   if (!cronartime){cronartime = "0 5,35 * * * *";}
   //Check value isn't garbage, if it is set to default
-  if (!cronvalidate.isValidCron(cronartime)) {
+  if (!cronvalidate.isValidCron(cronartime,{ seconds: true })) {
     logger.main.warn('CRON: Invalid CRON configuration in config file. Defaulting to: "0 5,35 * * * *" ')
     cronartime = "0 5,35 * * * *";
   } 
--- a/server/app.js
+++ b/server/app.js
@@ -25,6 +25,7 @@ var SQLiteStore = require('connect-sqlite3')(session);
 var flash    = require('connect-flash');
 
 
+
 process.on('SIGINT', function() {
     console.log( "\nGracefully shutting down from SIGINT (Ctrl-C)" );
     process.exit(1);
@@ -199,10 +200,16 @@ app.use(function(err, req, res, next) {
 // Add cronjob to automatically refresh aliases
 var dbtype = nconf.get('database:type')
 if (dbtype == 'mysql') {
+  const cronvalidate = require('cron-validator');
   // Get CRON from config
   var cronartime = nconf.get('database:aliasRefreshInterval');
   //If value is falsy (undefined, empty, null etc), set as default
   if (!cronartime){cronartime = "0 5,35 * * * *";}
+  //Check value isn't garbage, if it is set to default
+  if (!cronvalidate.isValidCron(cronartime)) {
+    logger.main.warn('CRON: Invalid CRON configuration in config file. Defaulting to: "0 5,35 * * * *" ')
+    cronartime = "0 5,35 * * * *";
+  } 
   var aliasRefreshJob = require('cron').CronJob;
   new aliasRefreshJob(cronartime, function() {
     var refreshRequired = nconf.get('database:aliasRefreshRequired')
--- a/server/app.js
+++ b/server/app.js
@@ -199,8 +199,9 @@ app.use(function(err, req, res, next) {
 // Add cronjob to automatically refresh aliases
 var dbtype = nconf.get('database:type')
 if (dbtype == 'mysql') {
+  var cronartime = nconf.get('database:aliasRefreshInterval');
   var aliasRefreshJob = require('cron').CronJob;
-  new aliasRefreshJob('0 5,35 * * * *', function() {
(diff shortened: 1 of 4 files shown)


geelongmicrosoldering:
side note, why does this cron exist for mysql only? something i'm missing?

server/config/default.json
```diff
@@ -11,7 +11,8 @@
   },
   "database": {
     "file": "./messages.db",
-    "type": "sqlite3"
+    "type": "sqlite3",
+    "aliasRefreshInterval": "0 1 3 * * *"
```
davidmckenzie, line 15: any reason to change the default from twice an hour?

davidmckenzie:
> side note, why does this cron exist for mysql only? something i'm missing?

It was a workaround for the poor alias refresh time that only seems to become a real problem when you get to the scale where you need mysql anyway, iirc. I don't think there was ever a reason _not_ to use it on sqlite from my vague memory, just that there wasn't a driver for it at the time.

server/config/default.json
marshyonline: Id also like this left at twice an hour by default

marshyonline (changes requested):
LGTM once default is changed back

server/config/default.json
geelongmicrosoldering: ah, that was a copy and paste error. default should be "0 5,35 * * * *", i believe i fixed it

geelongmicrosoldering:
> 
> It was a workaround for the poor alias refresh time that only seems to become a real problem when you get to the scale where you need mysql anyway, iirc. I don't think there was ever a reason _not_ to use it on sqlite from my vague memory, just that there wasn't a driver for it at the time.

> 
> MySQL Blocks updates to the capcode table while the refresh is running, preventing additions of capcodes. SQLite locks out the entire UI.
> 
> There is a few PR's and Issues with the thought processes in them i think.

Makes sense. I assumed it was there for a reason. And yeah sqlite locking out the UI was the primary reason i moved to mariadb. well, that and the large db size

geelongmicrosoldering:
@marshyonline sorry github is actually fairly foreign to me i don't fully understand the commit/review process yet. i just realised i can request a review of the change i made to fix the default cron back to twice per hour

server/app.js
```diff
@@ -199,8 +199,9 @@ app.use(function(err, req, res, next) {
 // Add cronjob to automatically refresh aliases
 var dbtype = nconf.get('database:type')
 if (dbtype == 'mysql') {
+  var cronartime = nconf.get('database:aliasRefreshInterval');
```
davidmckenzie, line 202: Sorry one other change - if the config item is not present (like in upgraded instances) this will fail, needs logic to fall back on a default

davidmckenzie (changes requested):

davidmckenzie (approved):
lgtm, thanks!

geelongmicrosoldering:
> lgtm, thanks!

wonderful!

I still don't understand github. PR's are fairly new to me. I feel like it wants me to click something somewhere..

davidmckenzie:
> I still don't understand github. PR's are fairly new to me. I feel like it wants me to click something somewhere..

it's just waiting on marshy's review then we can merge :)

geelongmicrosoldering:
> > I still don't understand github. PR's are fairly new to me. I feel like it wants me to click something somewhere..
> 
> it's just waiting on marshy's review then we can merge :)

ah, awesome that's cool I just didn't know if i did it right haha

DanrwAU (changes requested):
The only thing I can see is that it needs error handling for invalid values on the settings page as if you put gibberish in, it crashes the application on startup. 
Not sure which is easier, i'd almost be leaning towards GUI validation rather than a check serverside. 

Sorry to bounce this one back again, other than that it looks good.

DanrwAU:
@geelongmicrosoldering 

I fixed this one up, Added a node module to validate cron strings in app.js.

I also removed the "required" from the settings field as it would prevent users saving other settings until the default is pasted in.

geelongmicrosoldering:
> @geelongmicrosoldering
> 
> I fixed this one up, Added a node module to validate cron strings in app.js.
> 
> I also removed the "required" from the settings field as it would prevent users saving other settings until the default is pasted in.

Hey, thanks! I'm on my phone but looks good!

DanrwAU (dismissed):

DanrwAU:
@marshyonline @davidmckenzie if one of you 2 could look this one over. since i've now monkeyed with the code 🤣

geelongmicrosoldering:
> @marshyonline @davidmckenzie if one of you 2 could look this one over. since i've now monkeyed with the code 🤣

Actually worth querying, does null/empty/undefined validate with that validator?

If not you can probably remove my fallback line above because validating will revert the falsy value to default in your validation check anyway.

geelongmicrosoldering:
@DanrwAU Ah looks like that won't validate

cron.isValidCron('* * * * * *', { seconds: true });

If you pass Cron with seconds but don't tell the validator there's seconds it returns false according to docs

DanrwAU:
> @DanrwAU Ah looks like that won't validate
> 
> cron.isValidCron('* * * * * *', { seconds: true });
> 
> If you pass Cron with seconds but don't tell the validator there's seconds it returns false according to docs

Anyone giving it settings deserves it to fail. 

🤣  i'll leave it up to the group, but for this usecase i can't think of a single reason anyone would ever want to use seconds. Infact it may actually be detrimental as you could end up in a loop where it is running continuously if you set it too short?

geelongmicrosoldering:
> > @DanrwAU Ah looks like that won't validate
> > cron.isValidCron('* * * * * *', { seconds: true });
> > If you pass Cron with seconds but don't tell the validator there's seconds it returns false according to docs
> 
> Anyone giving it settings deserves it to fail.
> 
>  i'll leave it up to the group, but for this usecase i can't think of a single reason anyone would ever want to use seconds. Infact it may actually be detrimental as you could end up in a loop where it is running continuously if you set it too short?

The default has seconds lol, that was my point haha. "0 5,35 * * * *" is 6 parameters, we were already passing seconds

Translates to every 5m0s, 35m0s. Thats why it starts with 0, otherwise it would occur every second on the 5th and 35th minute

DanrwAU:
> > > @DanrwAU Ah looks like that won't validate
> > > cron.isValidCron('* * * * * *', { seconds: true });
> > > If you pass Cron with seconds but don't tell the validator there's seconds it returns false according to docs
> > 
> > 
> > Anyone giving it settings deserves it to fail.
> > i'll leave it up to the group, but for this usecase i can't think of a single reason anyone would ever want to use seconds. Infact it may actually be detrimental as you could end up in a loop where it is running continuously if you set it too short?
> 
> The default has seconds lol, that was my point haha. "0 5,35 * * * *" is 6 parameters, we were already passing seconds
> 
> Translates to every 5m0s, 35m0s. Thats why it starts with 0, otherwise it would occur every second on the 5th and 35th minute

Interesting.... it worked when i tested it 😕

I'll take a look when i get a sec

davidmckenzie:
fwiw, here's a thread re validation on the particular module we use: https://github.com/kelektiv/node-cron/issues/419

geelongmicrosoldering:
Sorry guys, I've had heaps on and totally temporarily abandoned this 8 months ago.

What was pending that i needed to fix?

@marshyonline @davidmckenzie

geelongmicrosoldering:
FWIW, can confirm that adding ` { seconds: true }` to cron validator, the behaviour is now as expected. before then, the custom cron was ignored because it didn't validate, so the default 5,35 was used instead.

Server now happy

marshyonline:
Ping for re-review plz

marshyonline (approved):
Looks ok to me

PR #474, opened 2021-06-03
Fix layout when message contains long word

geelongmicrosoldering:
Long words are bad.

# Description

There is an interesting result when a message contains a really long word. The page will infinitely get wider,and as a result all other messages that would have otherwise wrapped do not.

wordbreak_pm

The issue can be corrected by using word-break, and breaking a word that exceeds the width of its container. It shouldn't break other words, only those that are too long to fit. So there should be no visual change for anyone unless they encounter one of these messages.

wordbreak_pm_after

## Type of change

# How Has This Been Tested?

Tested on my own instance


DanrwAU (approved):
Looks good, this may fix some of the issues we've been having with i devices too

DanrwAU (approved):

PR #499, opened 2022-02-04
Automatic Alias Refresh for Specific Addresses

geelongmicrosoldering:
# Description

One of my current servers in production has over **1,000,000 messages**.
An Alias Refresh takes about **700-800 seconds**, during which the capcodes table is locked (mysql). This isn't so much due to hardware limitations, but seems to be because the refresh query performs a subquery for every message (so, over 1M queries, for which I don't have an index.) The structure of that query, and my database is probaby open for discussion, but i digress. I already [modified the alias refresh cron](https://github.com/pagermon/pagermon/pull/473)

In any case, 99% of the time, I'm adding an Alias with a specific address. That is, 1234567, not 1234___
Basically, I really only need to refresh the addresses for that specific address. If the address was previously part of a wildcard match, the other messages in the wildcard address match (say, 1234987) don't need to be refreshed.

So, I found myself manually refreshing messages every time I add an alias, using a a query like
```
update `messages` set `alias_id` = (select `id` from `capcodes` where `messages`.`address` like `capcodes`.`address` order by REPLACE(address, '_', '%') DESC limit 1) where `address` = '1234567';
```
This usually takes a **fraction of a second**, and applies immediate changes to any messages that match the alias, while leaving the others alone.

I found myself thinking, this should just be triggered when saving the alias, assuming the alias is not a wildcard match.
So, that's what I've done. Thought I'd share. It's 1am, and I'm not entirely sure this is the best way of going about it, but it seems to work well.

In testing, saving an alias still only takes a fraction of a second, except now, **thousands of matched messages immediately display the new alias.**

Just a single checkbox that triggers a refresh on messages that match your alias. Maybe needs rewording:
specific_refresh

## Type of change

# How Has This Been Tested?

I have this running on 2 server, both of which are running the latest release.

- [x] Running on Production Server with over 1M messages
- [x] Running on Production Server with over 100K messages

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.3.12 - TBD
+* Refresh alias automatically when single capcode alias is updated @geelongmicrosoldering
+
 # 0.3.11e - 2022-01-19
 * [HOTFIX ] Downgrade installed version of multimon-ng @marshyonline
 
--- a/server/routes/api.js
+++ b/server/routes/api.js
@@ -942,9 +942,36 @@ router.route('/capcodes/:id')
                   console.timeEnd('updateMap');
                 })
             } else {
-              if (!updateRequired || updateRequired == 0) {
-                nconf.set('database:aliasRefreshRequired', 1);
-                nconf.save();
+              //Check if we can refresh just this specific alias
+              var specificRefresh = nconf.get('global:SpecificAliasRefresh');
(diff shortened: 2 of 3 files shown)


davidmckenzie (approved):
I don't think this needs to be behind a config flag tbh - could be enabled for all. But looks good from my side regardless :)

davidmckenzie (approved):

marshyonline (approved):
Lgtm

PR #434, opened 2020-08-03
Testers needed: Handle FLEX group messages

eopo:
Rewrote major parts of reader.js
Put handling for different protocols into seperate functions
Modified FLEX handling to process new multimon group message format introduced with EliasOenal/multimon-ng@46c4a53eeb07b8e0f8e371897983ddea4b470997 - Awaiting EliasOenal/multimon-ng#155 response to see if done right.
Hopefully fixes #426

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change if you use a modified version of reader.js
- [ ] This change requires a documentation update

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated changelog.md with changes made

PR #517, opened 2022-05-29
Catch errors in Reader.js to prevent halting due to badly formatted pager messages

### LINKED ISSUE #346: TypeError: Cannot read property '1' of null
Crashed today with this message.  Running on Raspberry Pi.

```
readline.js:1032
            throw err;
            ^

TypeError: Cannot read property '1' of null
    at Interface.rl.on (/home/pi/pagermon/client/reader.js:104:48)
    at Interface.emit (events.js:189:13)
    at Interface._onLine (readline.js:290:10)
    at Interface._line (readline.js:638:8)
    at Interface._ttyWrite (readline.js:926:16)
    at Socket.onkeypress (readline.js:168:10)
    at Socket.emit (events.js:189:13)
    at emitKeys (internal/readline.js:424:14)
    at emitKeys.next (<anonymous>)
    at Socket.onData (readline.js:1022:36)
Signal caught, exiting!

User cancel, exiting...
```

CramShack:
# Description

As described in error #346  catches any errors during the parsing of pager messages, thus preventing any sudden exiting of the script.  I expect the issue is recurring due to unexpected characters in messages. I've had this personally happen in pager messages where there are line breaks.

Fixes # (issue)

Try/catch wrapped around the main line.match parsing section. Logs to console.error.

** Parsing itself should be expanded upon to accommodate unusually formatted pages. **

# How Has This Been Tested?

Ran the reader with the same messages fed through to replicate the cause of "TypeError: Cannot read property '1' of null" killing the script.


CramShack:
> > > 
> > 
> > 
> > I'd say pop them in another pull request, as its your contribution....
> > Also, in re to multi-line, better to use a JS solution rather than relying on the system to do new line replacements. Could maybe separate each line into multiple messages, but that may require more work than its worth.
> 
> The problem is that thats typically not how Linux piping works. The script executes every line individually. You could delay sending the message to see if theres another line, but then multiple messages sent together would end up as a single message. You would be better to remove the line break and execute reader.js once for one message, than treating every line as though there might be another coming. Especially when the line break will get stripped out by the time it makes it to the client anyway by the browser, because browsers dont render newlines in html anyway.

As I said, more work to refactor it for an uncommon use-case. 


Here, I was referring to using JS for stripping the newlines rather than using a linux command, 
> Also, in re to multi-line, better to use a JS solution rather than relying on the system to do new line replacements.

geelongmicrosoldering:
Js doesnt see the newlines is my point. Js never sees "message body \n more message body". Js just gets "process this", followed by "now process this".

Im not sure what the issue with not using JS is, given that the reader.js file is specifically made to read the output of a linux command (multimon-ng) anyway, which is being fed from another (rtl-fm). If you eliminate the non js commands, you also eliminate the need for the js.

SloCompTech:
> if (address.length > 2 && trimMessage.length > 1)

Here probably should be:

> if (address.length > 1 && trimMessage.length > 1)

since 9 is already valid pocsag address if I remember correctly

geelongmicrosoldering:
> > if (address.length > 2 && trimMessage.length > 1)
> 
> Here probably should be:
> 
> > if (address.length > 1 && trimMessage.length > 1)
> 
> since 9 is already valid pocsag address if I remember correctly

If 9 is valid then >1 wont catch it either.

I should have explained the 2 though. The erroneous messages from memory contained a single space or an arbitrary character, and were coming from an address of 99.

I actually wasnt aware that that was a valid address. Nonetheless, if you change it to >1 it will still filter the empty messages that were causing the issue. It wont filter the arbitrary characters, but thats what filters are for, so that works for me.

eopo:
I did already completely rewrite reader.sh in #434, but no-one that originally had the problem did participate in testing, and since it's an edge case ...

marshyonline:
Can we confirm this is still needed?
Drop back into non-wip if so plz :)

DanrwAU:
closed due to inactivity

Issue #519, opened 2022-06-13
Add support for non-standard port numbers for mysql databases


MitchDrage:
**Is your feature request related to a problem? Please describe.**
When attempting to deploy to docker hosts with more than on mysql database, unique port numbers are required.

**Describe the solution you'd like**
Addition of a new field in the settings page to support non-standard ports.

**Describe alternatives you've considered**
Not required, I believe.

**Additional context**
N/A

DanrwAU:
Fixed in #520

PR #520, opened 2022-06-13
My sql non standard portnumbers

MitchDrage:
# Description

Add support for non-standard port numbers for mysql databases, as is common when deploying docker containers.

Fixes # (issue)
https://github.com/pagermon/pagermon/issues/519

## Type of change

# How Has This Been Tested?

I haven't got a test environment, but I believe this change to be quite simple due to the use of Knex. I made these same changes to my running container and tested against a mysqldb running on 3307 (standard port is 3306)

# Checklist:


geelongmicrosoldering:
Looks good. Probably needs a `|| 3306` at nconf line or somewhere for people upgrading existing config where port is not defined. Idk

MitchDrage:
Good call.
Might need two actions though.
If we only add the `|| 3306` then I don't think it will make it's way in to a configuration file. If that's ok, then ignore the rest of this below and let me know to change that line...

Otherwise, if we only set a default value in the input in the settings file, connections from an upgraded server will fail unless someone goes back in to the settings page and re-saves their config.
So we would need both, one to allow upgraded servers to boot properly, and the second to set the default for the next time someone opens the settings page.
So, after I fix the fact that I forgot to rename the copied form class from database.server to databade.port...
Untested thus far, but something like this look ok? (Typing on my phone, and I can't seem to add the correct formatting here in a code block so excuse the lack of indentation)
`<script>`
  `  function setDefaultMysqlPort() {`
 `   document.getElementByName("database.port").defaultValue = "3306";`
`  }`
`</script>`
`window.onLoad  = setDefaultMysqlPort;`

eopo:
Hmm, it might be better to include some sort of migrations system that adds the default value to configuration, when it isn't present on startup

geelongmicrosoldering:
> Hmm, it might be better to include some sort of migrations system that adds the default value to configuration, when it isn't present on startup

Yeah i wondered about the best place for this to occur and how. Doing that would mitigate the need for the || 3306 as well as adding anything to the admin settings template to fix it.

You're probably right. Get rid of the this || that, the script above, and have a "if conf is mysql etc and database port not set, set it in config file" in app startup.

Ill throw a snippet together on Monday. @MitchDrage i see what you were going for and you are right. It should make its way to the config file if it isnt set. @eopo idea kills 2 bird and only has to run once so might be better.

eopo:
I'll hack something together that runs on startup, checks for a config version variable in config.js and runs migration scripts found in a separate dir as needed.

geelongmicrosoldering:
> I'll hack something together that runs on startup, checks for a config version variable in config.js and runs migration scripts found in a separate dir as needed.

i didnt see you write this but i got out of bed to do it while i remember haha

```
var dbtype = nconf.get('database:type');
// Set the database port if none found, for backwards compatibility
if (dbtype == 'pg' || dbtype == 'mysql' || dbtype == 'mssql') {
	if (!nconf.get('database:port')){
		nconf.set('database:port', 3306);
		nconf.save();
	}
}
```

can drop into app.js just under the `// set the theme if none found, for backwards compatibility` block. tested on my production server, seems to work

eopo:
Looks good, but I think we should make this a separate component and have every future change that needs something like config to be changed s separate file that gets loaded.

geelongmicrosoldering:
> Looks good, but I think we should make this a separate component and have every future change that needs something like config to be changed s separate file that gets loaded.

Probably yeah. I only did it app.js because someone else did it with theme lol. Maybe drop this in app.js for this PR and then you could do a PR that seperates this, theme, and other one time config file changes into a seperate component?

MitchDrage:
Great work everyone.
I'm honestly pretty new at this and appreciate all of the inputs to help solve the enhancement request.
I only really know a bit of python and docker, so anything js is a bit of an exercise in brain-expansion for me.

I've added new commits to the branch with the suggestions.

marshyonline (approved):
lgtm

eopo (approved):
Looks good for what it is.
I think we should move db configuration to ENV in the future though.

PR #530, opened 2022-08-11
Fix failing dependencies

eopo:
# Reason
The packaces `sqlite3` and `oracledb` are causing a lot of trouble when trying to run `npm install` on a lot of machines. For devices with apple silicon chips, there are no pre-built packages available, and manually building them does require some advanced knowledge

# Measures
Moved `sqlite2` and `oracledb` from `depencies` to `optional-dependencies`.
npm tries to install these, but doesn't fail the entire installation process, if one of these packages fails to install.
To make sure that the user has all the packages he needs to, I added a script that checks presence of the dependencies needed for the selected database type and aborts the run, if they are missing.

# Tests
Tests did run fine, except for user logout route, which always fails on my machine.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -67,6 +67,8 @@ if (azureEnable) {
              .start();
 }
 
+checkForDbDriver(nconf.get('database:type'));
+
 var dbinit = require('./db');
     dbinit.init();
 var db = require('./knex/knex.js');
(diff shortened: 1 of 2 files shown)


geelongmicrosoldering:
I like this. This is useful

geelongmicrosoldering:
i may know why your logout route fails too. there's an issue with a specific line that renders errors for express, that produces an error. i have it commented out on all of my machines or you get 5xx errors and have to manually go to logout url to remove the session. i think i raised an issue with it regarding cloudflare/nginx, but they dont seem to be at fault

SloCompTech (approved):

marshyonline:
LGTM!

marshyonline (approved):

Issue #500, opened 2022-02-12
Cloudflare + express causes 520 error.


geelongmicrosoldering:
**Describe the bug**
_I actually "fixed" this a while ago. Forgot how, reinstalled, and had to chase it again._

I'm not familiar enough with express/node to even begin to debug this, but commenting out the following fixes it on 3 servers:
`res.render(path.join(__dirname,'themes',theme, 'views', 'global', 'error'), { title: title });`
For someone familiar this probably isn't super difficult to squash.

Basically, while it was never officially supported to my knowledge, I use Cloudflare for my nameserver, CDN, etc. The setup is
`Client > Cloudflare DNS > NGINX Reverse Proxy > Node.js (pagermon)`
The issue comes when users login/logout, or when a session expires. Clients get an error 520 from Cloudflare. Depending where you look, you get a "headers already sent" error.

It basically turns out that the error is caused by the code, intended to render the error. Go figure.

For now, I'm just leaving the render disabled.

**To Reproduce**
1. Install Pagermon (any version)
2. Configure NGINX for reverse proxy
3. Configure domain to use Cloudflare DNS
4. Navigate to Domain
5. If you don't immediately get a 520 error, log in/out
6. Bypass the error by navigating to /auth/login directly, or comment out the res.render line in app.js

**Expected behavior**
Unauthorized clients, and clients logging out should be redirected to /auth/login

**Screenshots**
![cf520](https://user-images.githubusercontent.com/14150258/153693326-87c3e7dc-8795-470a-85fe-8787dee2b676.png)

**Environment information:**
 - OS: Ubuntu 20.04, although the same occurs on Rasbian, 18.04, and on multiple Pagermon releases

geelongmicrosoldering:
> This is an interesting one - I don't have this issue behind Apache Proxy + CF. Will need to have a deeper dive into this one.
> 
> Are you running secure mode or just default fresh install?

ah yep, should have noted that.

That's with secure mode enabled

crazyfella1:
I also am having a similar issue with Client > Cloudflare DNS > Cloudflare Argo Tunnel > Nodejs (pagermon). The work around heading direct to site /auth/login works fine. I will comment out the line highlighted above see if that makes a difference.

vk2mpj:
I too am experiencing this issue now. Applying the workaround seems to have gotten me back online

PR #549, opened 2023-03-16
Fix Error [ERR_HTTP_HEADERS_SENT]

nic-hanna:
# Description

Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client

Combine res.status(200) and res.send(message) to res.status(200).send(message)

# Type of Issue

Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Tested on local machine

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.3.12b - 2023-03-16
+* Fix [ERR_HTTP_HEADERS_SENT] error in /api route by combining res.status and res.send @nic-hanna
+
 # 0.3.12a - 2022-12-23
 * Revert #531 due to ongoing issues with crashes and 500 errors. @Danrw
  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.3.12b - 2023-03-16
+# TBA
 * Fix [ERR_HTTP_HEADERS_SENT] error in /api route by combining res.status and res.send @nic-hanna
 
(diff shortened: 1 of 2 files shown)


geelongmicrosoldering:
Ooh interesting. I dont have time to look into this right now but i feel like this is somehow related to #500

nic-hanna:
> Ooh interesting. I dont have time to look into this right now but i feel like this is somehow related to #500

I don't believe the issues would be related, this fix is only due to the usage of res.status in this code base as been deprecated and produces a warning not error.

geelongmicrosoldering:
See line 207 of app.js

I think you inadvertently solved an issue you werent looking for in a different file. I never notice the res.render was preceded by res.status

nic-hanna:
Might be worth going through all the routes and checking all the res.status.

geelongmicrosoldering:
Yeah i think so. It seems replacing 207/208 with
```res.status(err.status || 500).render(path.join(__dirname,'themes',theme, 'views', 'global', 'error'), { title: title });```

Would solve that issue. I never returned to it and seeing this PR jogged my memory.

DanrwAU:
Changed base repo to develop. We don't merge directly to master

DanrwAU (dismissed):
Looks good, matches express doco 
https://www.geeksforgeeks.org/express-js-res-status-function/?ref=ml_lbp

DanrwAU (approved):

PR #556, opened 2023-05-12
A new Plugin and a fix for the Telegram Plugin

alphaO4:
# Description

I did two things. 
First I created a completly new Plugin, called Meshtastics LoRa API.
This Plugin uses the [python CLI](https://meshtastic.org/docs/software/python/cli) from Meshtastic to send messages to certain chat ID's. This is extremly usefull for cases where there might not be a working cellular connection, since these Meshtastic devices work completle OffTheGrid. 

Dependencies: Meshtastics Python CLI

The Second thing is, that I updated the Telegram Plugin to be able to handle multible chat-id's.
This is usefull in case you don't want to create a Telegram group, or you want to send different Alarms to multible Telegram Users/Groups.
Both, Users and Groups can be given simultainiously. (Tho only in ID form)
Please note, that I disabled the Markdown in "sendMessage()", since it caused an escape character error with encrypted POCSAG messages, that had an escape character. 

Fixes # (issue)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] This change requires a documentation update

# How Has This Been Tested?

I tested this both under test conditions (with the ./sampleMessage.sh) and under real world continions. 
These changes are now running for a week in a "prod" enviroment, without any problems.

- [X] Using ./testMessage.sh to trigger aliases and therefore Messages
- [x] Letting it run in prod

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
-# 0.3.12b - 2023-05-12
-* Added a new Plugin [Meshtastics LoRaA PI](https://github.com/pagermon/pagermon/blob/master/server/plugins/MeshtasticsLoRaAPI.js) utilising the Meshtastic [Python CLI](https://meshtastic.org/docs/software/python/cli)
-* Added ability for Telegram Plugin to receive multiple chat/group IDs
+# Unreleased - 2023-05-15
+* Added a new Plugin [Meshtastics LoRaA PI](https://github.com/pagermon/pagermon/blob/master/server/plugins/MeshtasticsLoRaAPI.js) utilising the Meshtastic [Python CLI](https://meshtastic.org/docs/software/python/cli) #556 @AlphaO4
+* Added ability for Telegram Plugin to receive multiple chat/group IDs #556 @AlphaO4
 
 # 0.3.12a - 2022-12-23
 * Revert #531 due to ongoing issues with crashes and 500 errors. @Danrw
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.3.12b - 2023-05-12
+* Added a new Plugin [Meshtastics LoRaA PI](https://github.com/pagermon/pagermon/blob/master/server/plugins/MeshtasticsLoRaAPI.js) utilising the Meshtastic [Python CLI](https://meshtastic.org/docs/software/python/cli)
+* Added ability for Telegram Plugin to receive multiple chat/group IDs
+
 # 0.3.12a - 2022-12-23
 * Revert #531 due to ongoing issues with crashes and 500 errors. @Danrw
  
--- a/server/plugins/MeshtasticsLoRaAPI.js
+++ b/server/plugins/MeshtasticsLoRaAPI.js
@@ -18,7 +18,8 @@ function run(trigger, scope, data, config, callback) {
 
 
     // Build the command string
-    const command = `/home/pi/.local/bin/meshtastic --sendtext "${data.message}"  --ch-index ${data.pluginconf.MeshtasticsLoRaAPI.channelID}`;
+    // This will currently only work on Linux/Unix based machines.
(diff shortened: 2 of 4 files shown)


alphaO4:
This also closes the branch "[eature-telegram-multiple-targets](https://github.com/pagermon/pagermon/tree/feature-telegram-multiple-targets)"

DanrwAU:
We don't merge directly into master, please modify to merge into develop first.

server/plugins/MeshtasticsLoRaAPI.js
```diff
@@ -0,0 +1,40 @@
...
+    // Check if the channelID and message variables exist and have values
+    if (!data.pluginconf.MeshtasticsLoRaAPI.hasOwnProperty('channelID') || !data.hasOwnProperty('message') || !data.message) {
+      logger.main.error('Missing channelID or message variable.');
+      callback(data); // Pass the data to the callback function
+      return;
+    }
+
+
+    // Build the command string
+    const command = `/home/pi/.local/bin/meshtastic --sendtext "${data.message}"  --ch-index ${data.pluginconf.MeshtasticsLoRaAPI.channelID}`;
```
DanrwAU, line 21: This isn't going to work cross platform or even cross user accounts, needs to be fixed.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.3.12b - 2023-05-12
```
DanrwAU, line 1: Adding featurs would not be a B release. 

Releases are done by the team, add as Unreleased please, also follow the format below, add the PR number and @ tag of the author

server/plugins/MeshtasticsLoRaAPI.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "MeshtasticsLoRaAPI",
+    "description": "Integration for LoRa, utilising the provided HTTP API. <br>For it to work, you need to specify a chatURL file once on the Host. <br>To do this run: 'meshtastic --set-url https://meshtastic.com/o/thegivenurl  ' on the sender as well as the reciever. ",
```
DanrwAU, line 3: Need to add any pre-requisites the user requires as well, such as the python cli you are using.

DanrwAU (changes requested):
Few changes there @marshyonline @davidmckenzie can you take a look for anything i've missed :-)

geelongmicrosoldering:
We will also likely need to vigorously test the telegram plugin changes. Not because theres something wrong with your code, but because we actually had a multi destination telegram plugin and had to frantically revert it for.. reasons.

alphaO4:
Perfekt, I will make the changes as soon as I'm home from work.

alphaO4:
> We will also likely need to vigorously test the telegram plugin changes. Not because theres something wrong with your code, but because we actually had a multi destination telegram plugin and had to frantically revert it for.. reasons.

So far, I didn't have any problems, with regular Strings, of any length. 
What where the problems you guys encountered?

server/plugins/MeshtasticsLoRaAPI.js
alphaO4: How do you propose to make this crossplattform?
I thought about maybe adding a field in the settings asking for the path to Meshtastic. 
The problem with that tho, is that the user could give a path to any code, making the app vulnerable to a RCE.

Atleast for most Linux/Mac devices, you can use `${HOME}` or `~`. 
But adding support for Windows, could be quite complicated since the installation path could be vastly different for each user and used Python version.

DanrwAU: i don't have a solution for you off the top of my head, issue is currently we don't have any features gated behind an OS wall. (even the shell plugin is cross platform) so i don't think we'd want to go that route. @davidmckenzie @marshyonline  may want to chime in there tho

DanrwAU (changes requested):

DanrwAU:
Closing due to inactivity

PR #578, opened 2023-11-11
Chnages to npm 

Shaggs:
Just done an install on a raspberry pi

Encounter errors with installing pm2.

Updated for workaround


Shaggs:
Closing as this was not the issue.

I was installing on a raspberry pi and nodejs version was stuck on 6.

Will do an install script for rpi users as raspbian has js issues..

Will have another source for pushing data from a serial decoder.

Issue #579, opened 2023-11-12
Raspbain nodejs version 6


Shaggs:
As raspbain has its on nodejs versions pagermon doesn't behave as expected.

Pagermon need version 12.x or above but raspbain runs 6.x.

`apt-get update && apt-get upgrade -y` does not fix this issue.

I will upload a bash-script installer for raspbain.

This is also an issue for the pi image that is in the releases page (tired this also to make sure i wasnt dumb).

Shaggs:
Please assign and will submit a pull request after testing

PR #581, opened 2023-11-14
PDR scraper, R-Pi installer

Shaggs:
# Description
Add a scraper via serial for decoders. This will allow users with pager decoders to upload messages. Mulit-part messages are handled via the device so no need to mess around with that side of things.

The python does have some library's that need to be installed.
------------------
Raspberian  now has its own version of Nodejs installed being version 6. the install script will remove that version and install version 18 - 20 

Fixes # (issue)
Raspbain nodejs version 6 [#579](https://github.com/pagermon/pagermon/issues/579)
## Type of change

- [ V] Bug fix (non-breaking change which fixes an issue)
- [ V] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

New install of Raspbain to try install script.
PDR.py has been used on another project of mine for 3 years now with no issues.

# Checklist:

- [ V] My code follows the style guidelines of this project
- [ V] I have performed a self-review of my own code
- [ V] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/client/pdr.py
+++ b/client/pdr.py
@@ -1,0 +1,64 @@
+'''
+By S.Rees 2023 Python PDR Scraper.
+
+For use on pager decoder devices running from a serial device
+
+Edit as required'''
+
+import sys
+import os
+sys.path.append(os.path.realpath('.'))
+import serial
+import time
+Import requests
+from datetime import datetime
+import re
+import inquirer
+from serial.tools import list_ports
+questions = [
+    inquirer.List('device',
+                  message="Please pick a device",
+                  choices=[port.device+ ":"+ port.description 
+                  for port in list_ports.comports()],
+              ),
+]
+answers = inquirer.prompt(questions)
+dev = answers["device"].split(":", 1)
+ser = serial.Serial(dev[0], 19200, timeout=5)
+from clint.textui import puts, colored
+
+def post(flexcode, msg, when):
+		headers = {
+			'X-Requested-With': 'XMLHttpRequest',
+			'apikey': "", #pagermon APIKey
+			'User-Agent': 'PagerMon pdr.py',
+		}
+		params = {
+			"address": flexcode,
+			"message": msg,
+			"datetime": when,
+			"source": "",
+			}
+		requests.post('http://Your IP Address Here/api/messages', data=params, headers=headers)
+print("feed is now starting \nThis may take some time whist the signal is found")
+while 1:
+	line = ser.readline()
+	line = line.decode()
+	when = int(time.time())
+	if not line.strip():
+		pass
+	else:
+		try:
+			flexcode, sa, msg = line.split(' ', 2)
+			flexcode = flexcode[2:]
+			puts(colored.yellow(flexcode), newline=False)
+			puts(" [", newline=False)
+			puts(colored.green(when), newline=False)
+			puts("] ", newline=False)
+			if "CFSRES" in msg:
+				puts(colored.red(msg.strip()))
+			else:
+				puts(msg.strip())
+			post(flexcode, msg, when)
+		except Exception as e:
+			print(e)
--- a/server/package.json
+++ b/server/package.json
@@ -49,6 +49,7 @@
     "node-prowl": "latest",
     "node-uuid": "^1.4.8",
     "nodemailer": "latest",
+    "oracledb": "^4.2.0",
     "passport": "^0.3.2",
     "passport-local": "^1.0.0",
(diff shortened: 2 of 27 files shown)


client/pdr.py
```diff
@@ -0,0 +1,64 @@
...
+For use on pager decoder devices running from a serial device
+
+Edit as required'''
+
+import sys
+import os
+sys.path.append(os.path.realpath('.'))
+import serial
+import time
+Import requests
```
s3m1s0n1c, line 13: Line 13
Import requests

Should be

import requests

server/package.json
```diff
@@ -89,4 +90,4 @@
   "optionalDependencies": {
     "oracledb": "^4.2.0"
   }
-}
\ No newline at end of file
+}
```
s3m1s0n1c, line 93: serial-0.0.97 needs to be added to packages

client/pdr.py
```diff
@@ -0,0 +1,64 @@
...
+			'apikey': "", #pagermon APIKey
+			'User-Agent': 'PagerMon pdr.py',
+		}
+		params = {
+			"address": flexcode,
+			"message": msg,
+			"datetime": when,
+			"source": "",
+			}
+		requests.post('http://Your IP Address Here/api/messages', data=params, headers=headers)
```
s3m1s0n1c, line 42: Suggest pulling the following data from /client/config/default.json so its consistent with the rest of the project

* IP Addresss/Hostname
* API Key
* Source/identifier

s3m1s0n1c (commented):
Thanks for your PR @Shaggs 

See my suggestions I made in comments..

Will test it the next few days and give more feedback.

Don't think there will be many people who will have access to these devices..

I was lucky and managed to get a PSR3000 dirt cheap... 

Thanks
Sonic

client/pdr.py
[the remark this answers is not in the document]
Shaggs: No need for pm2 output. As this runs in parallel (start pm2 then run this in a sperate terminal window).

As this just pushes the data via post requests you will see the output in the terminal this is run from.

Happy to do a sperate document on how to make it run.

client/pdr.py
Shaggs: > Suggest pulling the following data from /client/config/default.json so its consistent with the rest of the project
> 
> * IP Addresss/Hostname
> * API Key
> * Source/identifier

When testing this was the json results pulled. When you test you will see these are the required arguments.

client/pdr.py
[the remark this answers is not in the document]
s3m1s0n1c: Oh yes @Shaggs, You are correct... Dunno what I was thinking..

client/pdr.py
s3m1s0n1c: What i'm trying to suggest is to parse the default.json and config.json file to get the Hostname, API Key & Source/Identifier instead of having to put it in pdr.py manually..

Shaggs: > What i'm trying to suggest is to parse the default.json and config.json file to get the Hostname, API Key & Source/Identifier from the config and use variables in pdr.py for each one instead of having to put it in pdr.py manually..

Happy to do so. I'll add it as an update (got a bit on)

Shaggs (commented):
will make some more changes when time permits. Those that use python shouldn't have too many issues.

Happy to answer any that come in until I can push updates

s3m1s0n1c:
@Shaggs 

I get the following error..

Guess the output of mine is different so unable to review it..

Traceback (most recent call last):
  File "pdr.py", line 46, in <module>
    line = line.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 2: invalid start byte

geelongmicrosoldering:
> Guess the output of mine is different so unable to review it..

Having worked with a few different hardware decoders, i can tell you that the output of all of them differ slightly. Its also usually programmable.

Some of them also output messages to a socket rather than UART. So you have to setup a socket listener on the server and then pipe the data somewhere.

For that reason i basically make the solution on the fly. Theres a few ways to approach that, but the easiest way for most would be to form a string that looks like the output of multimon and pipe it straight to reader.js

You'll have a hard time coming up with a one size fits all solution, even with hardware decoders from the same manufacturers, in my experience.

eopo:
I suggest not merging that into the main repo, but rather have it as a separate repo and link to it in the readme.

eopo:
Also, please always keep different features in their own respective branches and pull requests, so we can merge one and reject the other. ;)

PR #597, opened 2024-04-15
Feature better sockets

eopo:
# Description

Improve the whole socket handling:
- Sockets use authentication now!
- We don't use socket namespaces but socket rooms now
- We don't need to choose which socket namespace to listen on client side anymore
- On connection, server side assigns the client to a socket room according to its user role
- In message view, sockets always maintained the same amount of messages. If we only had two messages in that view, a new message would pop out the last one and place the new one. Now, we'll fill the page until the limit is reached and just then remove old messages.

## Type of change

# How Has This Been Tested?
- [X] Extensive testing on local client
- [X] Passed mocha tests

# Checklist:


DanrwAU:
Oooo shiny, i'll take a looksie

eopo:
Needs adapting to #599, if that is merged beforehand!

DanrwAU (approved):
approved to build test version

PR #598, opened 2024-04-16
Fix: View alias filter

eopo:
# Description
As of now, the message view isn't really precise when a filter is applied.
It does show, what value we're looking for, but not, what kind of value it is.
This PR implements that the filter value is preceded by the field name.
I'd really love to add the alias name, when we're searching for an alias, but Angular won't let me do it (Just kidding, I just don't know how).

## Type of change

# How Has This Been Tested?

- [X] Tests on local system
- [X] Passes mocha tests

# Checklist:


DanrwAU (approved):
approved to build test version

PR #617, opened 2024-10-15
Fix inconsistent tests

### LINKED ISSUE #476: Inconsistent Tests
https://github.com/pagermon/pagermon/blob/32a5c87e55a1f16aedf24f8ae24ec5633b22268b/server/test/routes.api.messages.id.test.js contains different requests to the api/messages/:id endpoint.
https://github.com/pagermon/pagermon/blob/32a5c87e55a1f16aedf24f8ae24ec5633b22268b/server/test/routes.api.messages.id.test.js#L45 wants the result of messages/:id to be an object
https://github.com/pagermon/pagermon/blob/32a5c87e55a1f16aedf24f8ae24ec5633b22268b/server/test/routes.api.messages.id.test.js#L69 wants the same result to be an array
https://github.com/pagermon/pagermon/blob/32a5c87e55a1f16aedf24f8ae24ec5633b22268b/server/test/routes.api.messages.id.test.js#L106 again wants it to be an array
In this case, object seems to make more sense, because we just really want one result.

Additionally, the file https://github.com/pagermon/pagermon/blob/32a5c87e55a1f16aedf24f8ae24ec5633b22268b/server/test/routes.api.messages.test.js should contain requests to the api/messages/ endpoint (Without :id)
However, in the line below, it does contain a request to an :id endpoint, that is already included in the other test file.https://github.com/pagermon/pagermon/blob/32a5c87e55a1f16aedf24f8ae24ec5633b22268b/server/test/routes.api.messages.test.js#L85

DanrwAU:
# Description

Fixes inconsistencies in tests as per #476 
Fixes #476
## Type of change

# How Has This Been Tested?

PR to run tests - see if all tests pass!

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# TBA
+* Fix #476 - Fix inconsistent tests. #617 @Danrw
+
 # 0.4.0 - 2023-06-04
 * Update Readme #552 @eopo
 * Add Ntfy.sh Plugin #570 @Danrw
--- a/server/routes/api.js
+++ b/server/routes/api.js
@@ -478,7 +478,7 @@ router.route('/messages/:id')
           if (pdwMode && !row.alias) {
             res.status(200).json({});
           } else {
-            res.status(200).json(row);
+            res.status(200).json(row[0]);
           }
         }
       })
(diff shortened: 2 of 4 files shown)


DanrwAU:
@eopo i remember why this was an issue now. 

We monkey with the result in the API that turns it into an object when we remove the capcode. If we don't make any changes, the raw row returned is an array with 1 object. Probably actually an API bug rather than a Test issue.

PR #624, opened 2024-12-12
Update SQLite3 Message Search query logic

### LINKED ISSUE #623: Improve Search in SQLite3 DBs
**Is your feature request related to a problem? Please describe.**
When performing a search, it will only search and return the result of the full word.
EG: In Australia (Specifically, Victoria), the SES will send a page that starts with **Hb** then immediately the job code **S123456789**. Follow up messages generally have the job code elsewhere in the message (as they're non-standardised) and will have a space before the _S_. When searching for these jobs, only the messages where **S123456789** are a standalone word are being returned as a result
For example, this SQL query (using _where message like '%S240850600%'_)
image
Only returns lines 2 & 3 when searched for in Pagermon
image

**Describe the solution you'd like**
Search function should return all results from the search, ignore if they are whole-words or not

**Describe alternatives you've considered**
No alternative

**Additional context**
I already have a fix for this, just generating the FR so I have something to generate the PR against

bullseye555:
# Description

Update from MATCH (whole word results only) to LIKE (string results, regardless of where the string is in the decode) for SQLite3 Message Search string

Fixes #623 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

- [X] Tested searching whole-words
- [X] Tested searching for partial words
- [X] Tested searching for numbers
- [X] Tested searching for strings containing symbols (eg: **@,#/*!$^&()-+.**)

##NOTE: SQL wildcard symbols **_** and **%** continue to function as wildcard symbols

# Checklist:


DanrwAU:
Rebased to develop

DanrwAU:
This bug exists on other database types also afaik. Any chance of a holistic fix across all DB types?

eopo:
We could switch from MATCH to LIKE in general. All three DBs systems currently supported would support LIKE, with the only (manageable) pinpoint being them not uniform with case sensitivity. We might want to have some way to make whole world search available though. I think it might be best to implement some sort of query language that enables searches as
`alias:"CITY" message:"FIRE" -message:"Test" `

PR #626, opened 2025-01-22
MySQL authentication support upgraded to the latest auth method (but backwards compatible)

HarmvandenBrink:
# Description

MySQL authentication support upgraded to the latest auth method. You can now choose an authentication method from a drop down menu when you select MySQL/MariaDB as a database target. This allows for backwards compatibility as well as future support for MySQL version > 8.0.

## Type of change

# How Has This Been Tested?

I installed a fresh Raspberry Pi and git cloned the repository with the new feature. Then I connected the pagermon server to a new MySQL server (version 9.1.0), with the SHA2 caching option. Next I connected to a old MySQL server which supported the native_mysql_password, this also went well. Both databases could be connected and user.

# Checklist:

### DIFF
--- a/client/reader.sh
+++ b/client/reader.sh
@@ -20,6 +20,7 @@
 ######################################################################
 # node reader.js : the cool thing that makes magic happen
 ######################################################################
-rtl_fm -d 0101 -E dc -F 0 -l 15 -A fast -f 148.5875M -s22050 - | 
-multimon-ng -q -b1 -c -a POCSAG512 -f alpha -t raw /dev/stdin | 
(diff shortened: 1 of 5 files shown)


DanrwAU:
Rebased to Develop - we don't merge directly into master.

client/reader.sh
DanrwAU, line 1: Why are we changing the client? This should not be included.

server/knexfile.js
DanrwAU, line 1: This is adding undocumented functionality - DB Connection pooling etc. 

@pagermon/collaborators does this need tests also?

DanrwAU (changes requested):

PR #634, opened 2025-04-17
Fruehjahrsputz 1/2: Callback Hell

### LINKED ISSUE #602: Searching for string AND other filters doesn't work
**Describe the bug**
Searching for a string and at the same time applying a filter does return all messages, that the string matches on, while disregarding the other filters.

**To Reproduce**
1. Visit messages view
2. Enter a search string
3. Manually add filter parameters in query string
4. All messages that match the string are shown

**Expected behavior**
Only messages, that meet all criteria should be shown.

**Screenshots**
Bildschirmfoto 2024-04-19 um 16 41 40

eopo:
# Description

In Germany, we have the beautiful word "Frühjahrsputz", which translates zu "Spring cleaning". It refers to getting rid of the winter's dirt in order to start the new year clean. This is usually done earlier than the end of April, but that's no problem - better late than never.

To ensure that pagermon makes progress in the long term, I have made a few plans. The current #Fruehjahrsputz consists of two parts:
- To increase the readability and maintainability of our code, I want to resolve the callbacks, which are heavily nested in many places. With “async/await”, JavaScript now offers a tried and tested means of keeping asynchronous code easy to read. In addition, there are now simpler functions from Express, Knex and the like at many points, so that some hacks that used to be necessary are no longer necessary.
- In a second step, I would like to continue updating our dependencies so that we can get rid of long overdue security vulnerabilities.

This pull request, which will be kept as a draft for the time being, should contain the measures of the first step. The development will be accompanied by ongoing tests. Despite the extensive scope of the changes, nothing should actually break.


RIKIKU:
This PR is massive! 
Untitled
I'm a little scared by the size of this PR but I'm pretty keen to see this succeed because I've been a little concerned by the security vulnerabilities.

PR #640, opened 2025-04-24
Update angular-textreplace.js

microsoldering1:
# Description

This is a total rewrite of angular-textreplace.js. It makes more sense (to me) and the performance difference is barely measurable (I measured it a fraction slower, but it was more functional, so it may have just been the extra regex that worked)

it processes everything in order, so the order of replacements/matches matters, and is processed exactly how it was entered.

# Fixes

It mitigates breaking tags/links by skipping matches that are inside tags and links, which prevents broken links.
For example. if you have a complex address match/replacement, and the address is 123 POLICE RD MYTOWN, but you also have a highlight on POLICE, previously one would break the other, leaving you with html displayed on the page and broken highlights/labels. Now, the POLICE in the address is ignored, because its already inside another link. of course, you can still do a regex replacement on the word inside the tag if you wish, because replacements are literal.

Another example would be highlighting both FIRE and STRUCTURE FIRE, where FIRE would break STRUCTURE FIRE by inserting a link inside a link

It gets rid of tokenizer and the .search method, because String.search seems to have a problem with regex when its more complex, and doing a search to see if the regex match is there before processing the match seems like double handling, seeing as String.replace will just return String if there is no match anyway

# How Has This Been Tested?

I actually tested it in March 2023 and am only doing the PR now, so I can't remember. I'm sure it will be fine. If someone wants to skim over it and see if they can perceive any issues when you get a chance, i commented it 

# Checklist:

### DIFF
--- a/server/themes/default/public/javascripts/angular-textreplace.js
+++ b/server/themes/default/public/javascripts/angular-textreplace.js
@@ -5,44 +5,49 @@ angular.module('angular-highlight', []).directive('highlight', function() {
 			attrs.highlightClass = 'angular-highlight';
 		}
 		
-		function arrSearch(nameKey, myArray){
-		    for (var i=0; i < myArray.length; i++) {
-		    	var rx = new RegExp(myArray[i].match, "gi");
-		        if (nameKey.search(rx) > -1) {
-		            return myArray[i];
-		        }
-		    }
-		}
+		//scope.highlight has the original message
+		//scope.replacement has all the replacements
 		
-		var rReplacer = function(match, item) {
-			var resultObject = arrSearch(match, scope.replacement);
+		//scope.replacement[i].highlight shows the mode
+		//scope.replacement[i].match contains the match
+		//scope.replacement[i].replace contains the replacement
+		
+		function rexReplacer(inputHtml) {
+			//Store the original HTML for later use
+			var html = inputHtml;
 			
-			//Fix issue with undefined variables, mostly from migrated configs.
-			if (typeof resultObject.highlight !== 'undefined') {
-				var thisMode = resultObject.highlight;
-			} else {
-				var thisMode = false;
-			}
+			//Perform this function for every Text/Regex Match
+			scope.replacement.forEach(function(textreplace) {
+			
+				//Fix issue with undefined variables, mostly from migrated configs.
+				if (typeof textreplace.highlight !== 'undefined') {
+					var thisMode = textreplace.highlight;
+				} else {
+					var thisMode = false;
+				}
+				
 
-			if (thisMode == "replace") {
-				var thisRex = new RegExp(resultObject.match)
-				var html = match.replace(thisRex,resultObject.replace);
-			} else if (thisMode) {
-				var html = '<a href="/?q='+match+'" data-toggle="popover" class="'+attrs.highlightClass+'" title="'+resultObject.replace+'" onmouseenter="$(this).tooltip(\'show\')">'+match+'</a>';
-			} else {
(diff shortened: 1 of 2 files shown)


server/themes/default/public/javascripts/angular-textreplace.js
```diff
@@ -5,57 +5,58 @@ angular.module('angular-highlight', []).directive('highlight', function() {
...
-			if (typeof resultObject.highlight !== 'undefined') {
-				var thisMode = resultObject.highlight;
-			} else {
-				var thisMode = false;
-			}
+			//Perform this function for every Text/Regex Match
+			scope.replacement.forEach(function(textreplace) {
+			
+				//Fix issue with undefined variables, mostly from migrated configs.
+				if (typeof textreplace.highlight !== 'undefined') {
```
eopo, line 23: Could be replaced with
```
const thisMode = textreplace.highlight || false;
```

`var` is outdated, `let` and `const` should be used for numerous reasons. If/else-declarations should be replaced with `let x; if (true) x = a; else x = b;`, with ternary operators or with the syntax shown above.

server/themes/default/public/javascripts/angular-textreplace.js
```diff
@@ -5,57 +5,58 @@ angular.module('angular-highlight', []).directive('highlight', function() {
...
-			} else if (thisMode) {
-				var html = '<a href="/?q='+match+'" data-toggle="popover" class="'+attrs.highlightClass+'" title="'+resultObject.replace+'" onmouseenter="$(this).tooltip(\'show\')">'+match+'</a>';
-			} else {
-				var html = '<a href="/?q='+match+'" data-toggle="popover" title="'+resultObject.replace+'" onmouseenter="$(this).tooltip(\'show\')">'+match+'</a>';
-			}
+				
+				if (thisMode == "replace") {
+					//Generate Regex for replacement
+					var thisRex = new RegExp(textreplace.match);
+					html = html.replace(thisRex,textreplace.replace);
```
eopo, line 34: Could use early return here to decrease indentation

server/themes/default/public/javascripts/angular-textreplace.js
```diff
@@ -5,57 +5,58 @@ angular.module('angular-highlight', []).directive('highlight', function() {
...
+				} else {
+					//Store match for use in label/highlight
+					var match = html.match(textreplace.match);
+					//Generate Regex for label/highlight, dont highlight if we are inside another html tag
+					var thisRex = new RegExp('(?<!<[^>]*)'+textreplace.match+'(?![^<]*<\/a>)');
```
eopo, line 39: Could use literal templating syntax here to improve readability

server/themes/default/public/javascripts/angular-textreplace.js
```diff
@@ -5,57 +5,58 @@ angular.module('angular-highlight', []).directive('highlight', function() {
...
+					if (thisMode) {
```
eopo, line 40: Could use ternary with template literals here:
```
html = html.replace(thisRex,`<a href="/?q=${match}" data-toggle="popover" ${thisMode? `class="${attrs.highlightClass}"` : ''} title="${textreplace.replace}" onmouseenter="$(this).tooltip(\'show\')">${match}</a>`)
```

eopo (dismissed):
Without having tested it, it looks quite good. Made some annotations regarding coding style, but nothing critical.

eopo:
Merged it into my custom branch, runs like a charm. Didn't have a situation with nested merges, but looking at the code i think this would be fixed by this pr.

DanrwAU (approved):
AI doesn't like this change, yolo

PR #633, opened 2025-04-14
Hotfix for logout route

eopo:
# Description
The /auth/logout route tried to access the user variable, after the user has already been logged out, just for logging purposes.
The username is now written into a variable, before the actual logout takes place, so it is available after 

## Type of change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [X] Ci-Tests

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated changelog.md with changes made

PR #651, opened 2025-08-16
Prevents ERR_HTTP_HEADERS_SENT when using QUIC

### LINKED ISSUE #650: ERR_HTTP_HEADERS_SENT when behind HTTP/3 Nginx reverse proxy
**Describe the bug**
When Pagermon is placed behind an Nginx reverse proxy that is running QUIC (HTTP/3) the HTTPS session is unable to be established by the web client and on the server side the following error is generated:
 `Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client`

**To Reproduce**
1. Enable QUIC on all sites on an Nginx reverse proxy.
2. Attempt to access pagermon via that proxy

**Expected behavior**
A HTTPS session over QUIC should be established and the site should function normally.

**Screenshots**

**Environment information:**
 - OS: docker on Ubuntu 24.04.3 LTS
 - Other details: The  implementation of Nginx is the following docker image `linuxserver/swag`

**Additional context**

RIKIKU:
# Description
When using an Nginx reverse proxy with QUIC (HTTP/3) enabled, the site will not load. The error message on the client side is an SSL error, however, the error message `Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client` is present on the server side. 

This PR solves the issue by adding a `return` after each response.  I'll be honest, I used AI to generate this fix. I've tested it locally and it appears to be working well. 

Why it happens with QUIC:
QUIC is a transport protocol, but most Node.js frameworks (like Express) are designed for HTTP/1.1 or HTTP/2. If you use a QUIC implementation that doesn't fully integrate with Express's response lifecycle, you might accidentally trigger multiple res.send(), res.end(), or header-setting calls.

Fixes #650 

## Type of change

# How Has This Been Tested?

I have tested this by building it in docker and running it in my production site and using all the functions I can think of.  It appears to be working. I have not used any of the automation tests. 

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # TBA
 * Fix PiBuilder @marshy #604
+* Fix Cannot set headers after they are sent to the client @RIKIKU #650
 
 # 0.3.13 - 2023-09-04
 * Add Config option to fix FA icon's no longer loading. @marshyonline
--- a/server/routes/admin.js
+++ b/server/routes/admin.js
@@ -57,8 +57,10 @@ router.route('/settingsData')
             fs.writeFileSync(confFile, JSON.stringify(req.body, null, 2));
             nconf.load();
             res.status(200).send({ 'status': 'ok' });
+            return;
         } else {
             res.status(400).send({ error: 'request body empty' });
+            return;
         }
     });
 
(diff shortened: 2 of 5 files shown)


eopo:
Redundant, is already included in #633.
For my understanding, it shouldn't be necessary to write out return for EVERY response, since most functions finish after the response is sent and there is no risk of trying to send twice, which triggers the error.

PR #654, opened 2025-11-27
User Password reset

Shaggs:
# Description

User password reset using SMTP settings from pagermon. This will email the user a temporary password that can be used to login.

Fixes # (issue)

Added templates, JS and updated files

- [X ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Tested on local device. New forgot password button added to logon.html

# Checklist:

- [X ] My code follows the style guidelines of this project
- [ X] I have performed a self-review of my own code
- [ X] I have commented my code, particularly in hard-to-understand areas


client/pdr.py
eopo, line 1: Please remove unrelated files from this PR

server/package.json
```diff
@@ -56,7 +57,7 @@
     "serve-favicon": "~2.4.2",
     "slack": "latest",
     "socket.io": "^2.0.2",
-    "sqlite3": "~4.0.2",
+    "sqlite3": "^5.1.7",
```
eopo: Does it still run the tests? Some sqlite changes are breaking!

server/routes/auth.js
```diff
@@ -2,11 +2,14 @@ const express = require('express');
...
 const router = express.Router();
 const bcrypt = require('bcryptjs');
+const crypto = require('crypto');
 const moment = require('moment');
 const nconf = require('nconf');
-
+const nodemailer = require('nodemailer');
 const confFile = './config/config.json';
-nconf.file({ file: confFile });
+const confDefault = './config/default.json';
```
eopo: Currently, if there is no config file, the default will be copied as config.json. Because of that i don't think there's a need to load the defaults again.
Actually, it is not necessary to re-setup nconf (giving it the config file etc.) per file, as it is already setup in `app.js`, which in turn requires all the other files - That makes them run in the same scope, so they have access to the already loaded nconf configuration. I don't blame you anyways, because our whole code does that too (We should remove that sometimes)

server/routes/auth.js
```diff
@@ -28,6 +31,47 @@ const lockoutCallback = function(req, res, next, nextValidRequestDate) {
...
 };
+const generateTempPassword = () =>
+        crypto
+                .randomBytes(9)
+                .toString('base64')
+                .replace(/[+/=]/g, '')
+                .slice(0, 12);
+
+const loadSmtpConfig = () => {
+        const smtpSettings = nconf.get('plugins:SMTP');
```
eopo, line 39: I get your point here, but a core feature should not rely on plugin configuration. 
It would be better to add a new mail section to the core config and make the smtp-plugin use that.

server/routes/auth.js
```diff
@@ -131,6 +175,63 @@ router.route('/logout').get(authHelper.isLoggedIn, function(req, res) {
...
+                if (!email) {
+                        return res.status(400).send({ status: 'failed', error: 'Email is required' });
+                }
+
+                const smtpConfig = loadSmtpConfig();
+
+                if (!smtpConfig) {
+                        return res.status(500).send({
+                                status: 'failed',
+                                error: 'SMTP is not configured. Please set the SMTP plugin details in Admin > Settings.',
```
eopo: The route should be accessible to users, but gives a hint directed to the admin. Although I don't think that this is security relevant in this case, I'd prefer giving a more general hint (Password reset not possible, please contact the administrator). I would wish for a log entry as well, so the admin can check the logs for why the reset failed.

server/routes/auth.js
```diff
@@ -131,6 +175,63 @@ router.route('/logout').get(authHelper.isLoggedIn, function(req, res) {
...
+                        });
+                }
+
+                try {
+                        const user = await db('users')
+                                .whereRaw('LOWER(email) = LOWER(?)', [email])
```
eopo, line 207: `.whereRaw('LOWER(??) = LOWER(?)', 'email', email)` so it uses the right ticks for all database drivers

server/routes/auth.js
```diff
@@ -131,6 +175,63 @@ router.route('/logout').get(authHelper.isLoggedIn, function(req, res) {
...
+                                .first();
+
+                        if (user) {
```
eopo, line 210: It's okay that way, but for legibility reasons I'd prefer an early exit like you did with missing smtp config:
```
if (!user) return ...
const tempPassword = ...
```

That removes one layer of indentation and for future edits, there's no need to check which if belongs to the else, because there's no else.

server/rpi-install.sh
eopo, line 1: Please remove unrelated files from PR, you're welcome to submit seperate PRs for more features

server/scripts/user-password-reset.js
eopo, line 1: Again, the whole project isn't a good example in that, but it from a DRY perspective, this is really redundant. I'd propose adding a `function resetPassword(email)` function, which is called by the router as well as the user-password-reset script. That way, you'd need no duplicated code and it would be easier to maintain the whole thing.

eopo (commented):
Looks really good to me. I've annotated a few things in the code.
I can't really say much for the frontend, because I really suck at frontend, so i've focussed on the backend code.

server/package.json
Shaggs: I have just done a fresh setup and seems to be running with no issues

server/routes/auth.js
Shaggs: Will look at this later today

server/routes/auth.js
Shaggs: I did it this way so it would get the built in config rather than another file or hard coding SMTP settings. 

Thought was easier just to pull it all from one place

server/routes/auth.js
Shaggs: I'll adjust the wording for this today. Play even go down the path of a checkbox in the plugin for user password reset

Shaggs:
Thanks. I will login later today and tidy up some of it and test it again to confirm all is working.

With the password reset comment I will "/change it to "Self service password resets are not configured".

I will be keeping the SMTP plugin as the source. This way two lots of SMTP settings don't have to be used. 

Thanks for the review.

Shaggs (commented):
Updates done to code as per comments. All changes that have been made have been tested prior

server/routes/auth.js
eopo: Yeah, and you're right with that. I just would do it vice versa. Don't let the core functionality (Password Reset) use a Plugin's config options, but let the Plugin use the core config options instead.

In my opinion, the following would be the best
```json
{
    ... configOptions,
    "db": { ... }
    "mail": {
        ... smtp settings
    },
    "plugins": {
        "smtp" : { ... }
    }
    ...
}
```

The password reset only ever uses the main 'mail' settings. These settings could be used for future stuff relying on email, support for API based mail services could be added etc.
The smtp plugin uses the plugin configuration. If there is none, it uses the main mail settings. (So there's no need to configure twice, if you only plan to use only one setting).

I think relying on the plugin configuration is technical debt that could bite us in the ass in the future. If we ever went to move plugin configurations to a different format, e.g. move it to the database, we'd have to pull that apart.

server/package.json
eopo: Okay, tests via github actions currently fail because they were pulled from the master branch, which still tries to use node 12. We'll have to re-try the current test suite for a usable result.

PR #657, opened 2026-06-24
Fix Shell plugin bash script support

### LINKED ISSUE #620: Bash script support in shell pluggin
Hi 
with the shell pluggin shell script work but bash ".sh" script dont work
is it possible to add support of bash language
thanks

### LINKED ISSUE #652: Trouble installing on Debian 13, Node 24
I'm trying to install pagermon on Debian 13. I'm following the guide (https://github.com/pagermon/pagermon/wiki/Server-Installation). When I get to the command `npm install` inside the `~/pagermon/server` directory, it throws a bunch of errors beginning with:

npm error code 1
npm error code 1
npm error path /home/user/pagermon/server/node_modules/sqlite3
npm error command failed
npm error command sh -c node-pre-gyp install --fallback-to-build

And it just bombs out from there. I made sure that the value for line 3 is correct: 
`"cwd" : "/home/user/pagermon/server"`. 

I've tried switching to Node 22 (LTS), verified that I ran every command in the guide above correctly. I ran `sudo apt-get install build-essential python3 pkg-config libsqlite3-dev` just to be sure that any dependencies were installed. I also ran the following commands to make sure:

`npm cache clean --force`
`npm install -g npm node-gyp node-pre-gyp`
`sudo apt-get install build-essential python3 libsqlite3-dev`
`rm -rf node_modules package-lock.json`

But running `npm install` still fails. I've attached the full error log. Any help would be greatly appreciated. Thanks!

[pagermon_npm_install_errors.txt](https://github.com/user-attachments/files/22000396/pagermon_npm_install_errors.txt)

ghost:
## Description

Linux Shell plugin scripts were executed through `/bin/sh`. Scripts using bash syntax such as `[[ ... ]]` or arrays therefore failed on systems where `/bin/sh` is not bash.

This PR runs Linux Shell plugin scripts with `bash`. Windows PowerShell behavior is unchanged.

## Testing

- Verified bash-specific `[[ ... ]]` syntax executes successfully.
- Diff is limited to the Shell plugin and changelog.

Fixes #620

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 0.4.1 - TBA
 * Update sqlite3 dependencies for current Node.js and Debian releases #652
+* Run Shell plugin scripts with bash on Linux #620
 * Package.json now used as source of truth for all version statements. #646 @Danrw
 * Fix critical security and functional issues in develop branch @Danrw
 * Fix authentication bypass vulnerability in API routes (req.isAuthenticated missing parentheses) @Danrw
--- a/server/plugins/Shell.js
+++ b/server/plugins/Shell.js
@@ -32,7 +32,7 @@ function run(trigger, scope, data, config, callback) {
         if(process.platform === "win32"){
           var child = spawn("powershell.exe", [file_path+full_file_name, '"'+data.address+'"', "@'\r\n"+data.message+"\r\n'@", "@'\r\n"+JSON.stringify(data)+"\r\n'@"]); //
         }else{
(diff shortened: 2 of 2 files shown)


eopo:
We generally merge to develop, not to master. I've changed the base branch, which resulted in some merge conflicts you'll unfortunately have to solve.
On a side node: Please try to keep separate features/fixes in seperate pull requests, so we can handle them independently.

eopo:
Would be a breaking change though - `bash` and `sh` are not the same, and there might be sh-scripts not running with bash. We should try to respect the shebang if available and default to sh in order to remain backwards compatibility

PR #659, opened 2026-06-24
Fix sqlite3 installation on current Node.js releases

### LINKED ISSUE #652: Trouble installing on Debian 13, Node 24

ghost:
## Description

Updates the SQLite dependencies so PagerMon installs on Debian 13 and current Node.js releases without falling back to the obsolete sqlite3 4.x build path.

Changes:
- sqlite3 5.1.7
- connect-sqlite3 0.9.16
- add server/package-lock.json for reproducible installs
- document Node.js 18+ and native-module build prerequisites

## Testing

- npm ci succeeds on Node.js 22.22.2
- sqlite3 5.1.7 loads and completes an in-memory query

Fixes #652


.gitignore
```diff
@@ -45,8 +45,7 @@ node_modules/
 server/node_modules/
 client/node_modules/
 jspm_packages/
-package-lock.json
-server/package-lock.json
+/package-lock.json
```
eopo, line 48: We can keep that for now, but need to discuss on project level if we want to commit lock files as they are meant to be committed.

eopo (approved):
LGTM

